PR #7, opened 2017-02-24
DR-922 - Send missing exit code when exiting the app + Refactor protractor configuration

gastonmancini:
### Background

Travis CI was succeeding even if the protractor tests fail on SauceLabs.

### Change

- Send the exit code on the close event. 
- Refactor protractor configuration.

### Pending to be done

- Revisit the SauceLabs badge that is showing "unknown" state. This error might be related wite the bug mentioned on [this comments](https://wiki.saucelabs.com/display/DOCS/Using+Status+Badges+and+the+Browser+Matrix+Widget+to+Monitor+Test+Results).


src/gulpfile.js
```diff
@@ -422,7 +422,7 @@ gulp.task('test:travis', function (done) {
   child_process
     .spawn(getNodeBinary(), argv, { stdio: 'inherit' })
     .on('error', function (e) { throw e; })
-    .on('close', function (e) { process.exit(); });
+    .on('close', function (e) { process.exit(e); });
```
andresmoschini: Great! could you appy the same fix to `test:component`?

andresmoschini:
Nice work with the refactoring @gastonmancini !!

Does it pass the tests? in that case :shipit: !

src/gulpfile.js
gastonmancini: Done!

arzafran:
approved

gastonmancini:
Thanks @arzafran!

PR #9, opened 2017-03-10
Dr-861 Domain Manager

mvillaseco:
Changes:
 - Add Test.
 - Add Service, Controller and view to add Domain Manager.

Reminder: This it's a WIP story.

Pending: 

* Confirm Backend API (I think that maybe i'll have to handle in a different way the response by the backend, but i'm not sure yet)
* Change the fake call to the real api call in the setting's service.
* Apply style/design/ux
* More tests


andresmoschini:
> Reminder: This it's a WIP story.

Could you detail what things are missed?

src/spec/page-objects/settings-page.js
```diff
@@ -0,0 +1,23 @@
...
+    this._confirmAddDomainButton = $('.domain--input-container button');
+    this._domainInputContainer = $('.domain--input-container');
+  }
+
+  getUrl(){
+    return this._url;
+  }
+  clickInputToggler(){
+    return this._togglerDomainInputButton.click();
+  }
```
andresmoschini: What do you think about to reduce the scope of this PR to a ReadOnly list of domains? after that you could add a new PR for Add, Remove, etc

mvillaseco:
I think that maybe i'll have to handle in a different way the response by the backend, but i'm not sure yet. And also later i'll have to change the fake call to the real api call in the setting's service.

src/spec/settings_spec.js
```diff
@@ -0,0 +1,37 @@
...
+  function beginAuthenticatedSession() {
+    browser.addMockModule('descartableModule', () => angular
+      .module('descartableModule', [])
+      .run((jwtHelper, auth) => {
+        var permanentToken = 'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYmYiOjE0ODQ2MzAzMTgsImV4cCI6MTQ4NzIyMjMxOCwiaWF0IjoxNDg0NjMwMzE4LCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjM0NzUxIiwic3ViIjoxMDAzLCJ1bmlxdWVfbmFtZSI6ImFtb3NjaGluaUBtYWtpbmdzZW5zZS5jb20iLCJyZWxheV9hY2NvdW50cyI6WyJhbW9zY2hpbmktbWFraW5nc2Vuc2UiXSwicmVsYXlfdG9rZW5fdmVyc2lvbiI6IjEuMC4wLWJldGE1In0.dQh20ukVSCP0rNXMWBh2DlPQXbP0uTaYzadRDNPXECI9lvCsgDKNXc2bToXAUQDeXw90kbHliVF-kCueW4gQLPBtMJOcHQFv6LfgspsG2jue2iMwoBC1q6UB_4xFlGoyhkRjldnQUV0oqBTzhFdXuTvQz53kRPiqILCHkd4FLl4KliBgdaDRwWz-HIjJwinMpnv_7V38CNvHlHo-q2XU0MnE3CsGXmWGoAgzN7rbeQPgI9azHXpbaUPh9n_4zjCydOSBC5tx7MtEAx3ivfFYImBPp2T2vUM-F5AwRh7hl_lMUvyQLal0S_spoT0XMGy8YhnjxXLoZeVRisWbxBmucQ';
+        auth.saveToken(permanentToken);
+      }));
+  }
+
+  it('should trigger correctly to show the add domain input', () => {
```
andresmoschini: I am not sure to understand this test name.

Are you verifying that `Settings page should show add domain input when you click in add domain button`?

What _correctly_ means in this case?

src/spec/settings_spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+    //Assert
+    expect(settings.getDomainInputContainer().getAttribute('class')).not.toContain('active');
+    expect(settings.getDomainInputContainer().getCssValue('height')).toEqual('0px');
+
+    // Act
+    settings.clickInputToggler();
+
+    // Assert
+    expect(settings.getDomainInputContainer().getAttribute('class')).toContain('active');
```
andresmoschini: I think that these implementation details could be inside of page object.

Also, did you verify if it is possible to use a high level indicator in place of a DOM or CSS one? Did you try with something like _visible_?

src/spec/settings_spec.js
andresmoschini: Also, what do you think about to remove add domains functionality from this PR?

src/wwwroot/partials/settings/domain-manager.html
```diff
@@ -0,0 +1,43 @@
...
+      <table>
+        <tr>
+          <th class="column-large">{{ 'domain_manager_name_th' | translate}}</th>
+          <th class="column-small">{{ 'domain_manager_status_th' | translate}}</th>
+          <th class="column-small">{{ 'domain_manager_actions_th' | translate}}</th>
+        </tr>
+        <tr class="column-records" ng-repeat="d in vm.domains">
+          <td>{{d.name}}</td>
+          <td>{{d.status}}</td>
+          <td><i>X</i> - <i>0</i> - <i>P</i></td>
```
andresmoschini: What are these values?

andresmoschini:
It seems that it is more mature than I though, nice! I will review it in deeper later.

src/wwwroot/controllers/DomainManagerCtrl.js
```diff
@@ -0,0 +1,46 @@
...
+      'settings'
+    ];
+
+    function DomainManagerCtrl($scope, settings) {
+      var vm = this;
+      vm.regexDomain = "(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{0,62}[a-zA-Z0-9]\\.)+[a-zA-Z]{2,63}$)";
+      activate();
+
+      function activate() {
+        loadUserDomains();
```
andresmoschini: Some time ago we have been talking about to try to make `activate` function to return promises, it could be useful in different scenarios, and in general it is not more complex, could you try in this case?

src/wwwroot/controllers/DomainManagerCtrl.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+
+      vm.toggleDomainInput = function(){
```
andresmoschini: The name of this function is _toggle..._ but it is only _turning on_

src/spec/settings_spec.js
mvillaseco: It seems that the function IsDisplayed() in protactor only works with visibility:hidden, we are currently using height:0 to hide the container in order to apply a transition after, i think we are going to need to keep the test this way.

PD: Sure, i'll try to place it inside the page object file

src/wwwroot/partials/settings/domain-manager.html
mvillaseco: Still in progress icons :(

mvillaseco:
The values are being translated in the UI by the "translate" method.

andresmoschini:
> The values are being translated in the UI by the "translate" method.
👍

andresmoschini:
❤️

andresmoschini:
❤️ ❤️ ❤️

andresmoschini:
Lovely work @mvillaseco !

PR #12, opened 2017-03-17
DR-957 Fix issues with big data user and small fixes

mvillaseco:
Changes:
 - Use class instead of directive in css selector.
 - Fix C3 font-size.
 - Fix too small date.
 - Fix API y settings connections view.
 - Full width button in templates


andresmoschini:
👍

PR #14, opened 2017-03-22
DR-963 Correct markup, uncomment add domain functionality and fix responsive issues

mvillaseco:
Changes:
 - Add test for Add Domain.
 - Uncomment Add Domain functionality.
 - Css for Responsive Add Domain Layout

### DIFF
--- a/src/spec/settings_spec.js
+++ b/src/spec/settings_spec.js
@@ -6,6 +6,8 @@ describe('Settings Page', () => {
     browser.removeMockModule('descartableModule');
     browser.removeMockModule('descartableModule2');
     browser.removeMockModule('descartableModule3');
+    browser.removeMockModule('descartableModule4');
+    browser.removeMockModule('descartableModule5');
   });
 
   function beginAuthenticatedSession() {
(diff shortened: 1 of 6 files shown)


src/spec/settings_spec.js
```diff
@@ -16,25 +16,31 @@ describe('Settings Page', () => {
         auth.saveToken(permanentToken);
       }));
   }
-  ///This test will be commented until we add the new Domain Functionality
-  //it('should show add domain input when you click in add domain button', () => {
-    // Arrange
-    //beginAuthenticatedSession();
-    //var settings = new SettingsPage();
+  it('should show add domain input when you click in add domain button', () => {
```
andresmoschini: I do not understand what the word `input` mean in this context.

Maybe it is referring to the text added, but it makes me think in a input element.

What do you think about to simply remove that word?

src/spec/settings_spec.js
```diff
@@ -16,25 +16,31 @@ describe('Settings Page', () => {
...
+      }));
+    var settings = new SettingsPage();
 
     //Act
-    //browser.get('/#/settings/domain-manager');
+    browser.get('/#/settings/domain-manager');
 
     //Assert
-    //expect(settings.isDomainInputVisible()).toBe(false);
+    expect(settings.isDomainInputVisible()).toBe(false);
```
andresmoschini: Since it is not the Assertion (in uppercase of this tests) I would remove the comment `// Assert`. I mean, since it is only to validate our setup, in my opinion it is part of the arrange.

Also `browser.get('/#/settings/domain-manager');` is part of Arrange section.

src/spec/settings_spec.js
andresmoschini: Oh, I understand now. what do you think about 
"should show `add domain` input when you click in add domain button"?

src/spec/settings_spec.js
```diff
@@ -59,8 +65,63 @@ describe('Settings Page', () => {
...
+    beginAuthenticatedSession();
+    browser.addMockModule('descartableModule2', () => angular
+      .module('descartableModule2', ['ngMockE2E'])
+      .run($httpBackend => {
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/domains/).respond(200, {
+          "domains": [{name: "relay.com"}, {name: "fromdoppler.com", disabled: true }, {name: "makingsense.com", disabled: true }],
+          "defaultDomain": "relay.com"
+        });
+      }));
+      browser.addMockModule('descartableModule3', () => angular
```
andresmoschini: I think that it could be declared inside `descartableModule2`

src/spec/settings_spec.js
```diff
@@ -59,8 +65,63 @@ describe('Settings Page', () => {
...
+          });
+        }));
+
+      var settings = new SettingsPage();
+
+      browser.get('/#/settings/domain-manager');
+      settings.clickInputToggler();
+
+      //Assert
+      expect(settings.isDomainInputVisible()).toBe(true);
```
andresmoschini: I think that this _expect_ is part of the _Arrange_ section, not _Assert_ one.

src/spec/settings_spec.js
```diff
@@ -59,8 +65,63 @@ describe('Settings Page', () => {
...
+      // Act
+      settings.setDomain('test.com');
+      settings.submitAddDomain();
+
+      // Assert
+      expect(settings.isDomainInputVisible()).toBe(false);
+
+  });
+
+  it('should show an empty list of domains', () => {
```
andresmoschini: Is this test duplicated?

src/wwwroot/services/settings.js
```diff
@@ -13,20 +13,27 @@
...
-      return deferred.promise;
-    }*/
+      return $http({
+        actionDescription: 'action_adding_domain',
+        method: 'PUT',
+        url: url
+      })
+      .then(function (response) {
+        return response;
+      });
```
andresmoschini: This `then` is not necessary

andresmoschini:
I did not check CSS and HTML, but the rest looks really nice. Check my opinion about `// Arrange`, `// Act` and `// Assert` marks, I think that, in general should be only one by a test, if my opinion is not clear or you are not agree, please lets talk about it to be in the same page.

andresmoschini:
Nice!, please wait for the test before merge.

PR #15, opened 2017-03-23
DR-965 Set Domain Default

mvillaseco:
Changes:
 - Add call to service.
 - Add content.
 - Add html for make default domain.

WIP: 
 - Confirmation text or icon for setting the default Domain.

### DIFF
--- a/src/wwwroot/services/settings.js
+++ b/src/wwwroot/services/settings.js
@@ -42,7 +42,9 @@
       return $http({
         actionDescription: 'action_setting_default_domain',
         method: 'PUT',
-        data: domain,
+        data: {
+          'name': domain
+        },
         url: url
       });
     }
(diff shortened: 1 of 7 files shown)


src/wwwroot/services/settings.js
```diff
@@ -32,7 +33,21 @@
...
+    function setDefaultDomain (domain) {
+      var url = RELAY_CONFIG.baseUrl
+        + '/accounts/' + auth.getAccountName()
+        + '/domains/'
+        + '/defaultDomain';
+
+      return $http({
+        actionDescription: 'action_setting_default_domain',
+        method: 'PUT',
+        data: domain,
```
andresmoschini: I think that data will have the follow format:
```
{
    "name": domain
}
```

andresmoschini:
Nice work @mvillaseco !

What do you think about to add the tests before merge it?

PR #17, opened 2017-03-28
DR-967 - Activate domain

mvillaseco:
Changes:
 - Add Activate functionality.
 - Refactor Add domain call to use the same as Activate.
 - Add Activate Test.

### DIFF
--- a/src/spec/page-objects/settings-page.js
+++ b/src/spec/page-objects/settings-page.js
@@ -9,7 +9,7 @@ class SettingsPage {
     this._domainInputButton = $('.domain--default-container button');
     this._domainsSetDefaultButtonList = element.all(by.css('.domain--default-action'));
     this._defaultDomain = $('.domain--default-container .default-domain');
-    this._domainsDisabled = element.all(by.css('.domain--disabled-action'));
+    this._disabledDomains = element.all(by.css('.domain--disabled-action'));
 
   }
 
@@ -59,10 +59,10 @@ class SettingsPage {
     });
   }
   getDisabledDomain(){
-    return this._domainsDisabled.getText();
+    return this._disabledDomains.getText();
   }
-  clickDisabledDomain(){
-    return this._domainsDisabled.then(function(val){
+  clickFirstDisableDomainButton (){
+    return this._disabledDomains.then(function(val){
       return val[0].click();
     });
   }
--- a/src/spec/page-objects/settings-page.js
+++ b/src/spec/page-objects/settings-page.js
@@ -9,6 +9,7 @@ class SettingsPage {
     this._domainInputButton = $('.domain--default-container button');
     this._domainsSetDefaultButtonList = element.all(by.css('.domain--default-action'));
     this._defaultDomain = $('.domain--default-container .default-domain');
+    this._domainsDisabled = element.all(by.css('.domain--disabled-action'));
 
   }
 
@@ -54,6 +55,14 @@ class SettingsPage {
   }
   clickSetAsDefault(){
     return this._domainsSetDefaultButtonList.then(function(val){
+      return val[0].click();
+    });
+  }
+  getDisabledDomain(){
+    return this._domainsDisabled.getText();
+  }
+  clickDisabledDomain(){
+    return this._domainsDisabled.then(function(val){
       return val[0].click();
     });
   }
--- a/src/spec/page-objects/settings-page.js
+++ b/src/spec/page-objects/settings-page.js
@@ -9,7 +9,7 @@ class SettingsPage {
     this._domainInputButton = $('.domain--default-container button');
     this._domainsSetDefaultButtonList = element.all(by.css('.domain--default-action'));
     this._defaultDomain = $('.domain--default-container .default-domain');
-    this._disabledDomains = element.all(by.css('.domain--disabled-action'));
+    this._activateButtons = element.all(by.css('.domain--activate-action'));
 
   }
 
@@ -58,11 +58,11 @@ class SettingsPage {
       return val[0].click();
     });
   }
-  getDisabledDomain(){
-    return this._disabledDomains.getText();
+  countActivateButtons(){
+    return this._activateButtons.count();
   }
-  clickFirstDisableDomainButton (){
-    return this._disabledDomains.then(function(val){
+  clickFirstActivateButton (){
+    return this._activateButtons.then(function(val){
       return val[0].click();
     });
   }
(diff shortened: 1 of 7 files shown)


src/spec/page-objects/settings-page.js
```diff
@@ -9,6 +9,7 @@ class SettingsPage {
     this._domainInputButton = $('.domain--default-container button');
     this._domainsSetDefaultButtonList = element.all(by.css('.domain--default-action'));
     this._defaultDomain = $('.domain--default-container .default-domain');
+    this._domainsDisabled = element.all(by.css('.domain--disabled-action'));
```
andresmoschini: What do you think about `_disabledDomains`?

andresmoschini: No, sorry, is a collection of buttons/links? what do you think about `_disableDomainButtons`?

src/spec/page-objects/settings-page.js
```diff
@@ -57,5 +58,13 @@ class SettingsPage {
       return val[0].click();
     });
   }
+  getDisabledDomain(){
```
andresmoschini: What does it means? `_domainsDisabled` is a collection, what is the result of `getText`?

src/spec/page-objects/settings-page.js
```diff
@@ -57,5 +58,13 @@ class SettingsPage {
       return val[0].click();
     });
   }
+  getDisabledDomain(){
+    return this._domainsDisabled.getText();
+  }
+  clickDisabledDomain(){
+    return this._domainsDisabled.then(function(val){
```
andresmoschini: does it click in all disable buttons?

andresmoschini:
Nice! Do the tests pass?

src/spec/page-objects/settings-page.js
mvillaseco: Right now i'm mocking only 1 disable button, should we consider to test when having more than 1?

andresmoschini: No no, but, I think that the method name should reflect what it is happening, because, in the future someone could add a new test scenario with more elements, and in this case, how it will work?

Does `clickFiirstDisabledDomainButton` be more precise? or I am missing something?

mvillaseco: Nop, that will be more precise :) thanks! i'll change it

src/spec/page-objects/settings-page.js
mvillaseco: Returns all the activate buttons

andresmoschini: And what is `['Activate']`? I mean, how do you name this method to be precise? maybe `getDisableDomainButtonTexts()`?

I think that maybe you can create a new method `getDisableButtonsCount()` in replace of this one, I think that it could be easy to understand.

andresmoschini: or `countDisableButtons()`

mvillaseco:
Yes :D

src/spec/settings_spec.js
```diff
@@ -132,4 +132,29 @@ describe('Settings Page', () => {
...
+        $httpBackend.whenPUT(/\/accounts\/[\w|-]*\/domains/).respond(201, {
+          "result": true
+        });
+      }));
+    var settings = new SettingsPage();
+
+    //Act
+    browser.get('/#/settings/domain-manager');
+    var disabledDomain = settings.getDisabledDomain();
+    settings.clickDisabledDomain();
```
andresmoschini: I think that the first two lines are part of the _preparation_, like this:

// Arrange
// . . .
browser.get('/#/settings/domain-manager');
var disabledDomain = settings.getDisabledDomain();

// Act
settings.clickDisabledDomain();

src/spec/settings_spec.js
```diff
@@ -132,4 +132,29 @@ describe('Settings Page', () => {
     expect(settings.getDefaultDomain()).not.toEqual(defaultDomain);
     expect(settings.getDefaultDomain()).toEqual("fromdoppler.com");
   });
+
+  it('should set as enabled domain correctly', () => {
```
andresmoschini: Maybe it should be `should disable domain`?

mvillaseco: The test it's for activate a Domain disabled, maybe "Should activate the domain correctly"=

andresmoschini: In that case, there is something that I am not understanding, please call me when you can.

src/spec/settings_spec.js
```diff
@@ -132,4 +132,29 @@ describe('Settings Page', () => {
...
+    // Arrange
+    beginAuthenticatedSession();
+    browser.addMockModule('descartableModule2', () => angular
+      .module('descartableModule2', ['ngMockE2E'])
+      .run($httpBackend => {
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/domains/).respond(200, {
+          "domains": [{name: "relay.com"}, {name: "fromdoppler.com" }, {name: "makingsense.com", disabled: true }, {name: "makingsense12.com" }],
+          "defaultDomain": "relay.com"
+        });
+        $httpBackend.whenPUT(/\/accounts\/[\w|-]*\/domains/).respond(201, {
```
andresmoschini: Is it possible to also test the request body? It could be useful to understand the test and for validate if the system is doing what we expect

mvillaseco:
Can i merge this? @andresmoschini  @nreal

andresmoschini:
I think so, @nreal could you review at least @mvillaseco's tests and verify if he is requesting API as you expect?

src/spec/settings_spec.js
```diff
@@ -132,4 +132,29 @@ describe('Settings Page', () => {
...
+    // Arrange
+    beginAuthenticatedSession();
+    browser.addMockModule('descartableModule2', () => angular
+      .module('descartableModule2', ['ngMockE2E'])
+      .run($httpBackend => {
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/domains/).respond(200, {
+          "domains": [{name: "relay.com"}, {name: "fromdoppler.com" }, {name: "makingsense.com", disabled: true }, {name: "makingsense12.com" }],
+          "defaultDomain": "relay.com"
+        });
+        $httpBackend.whenPUT(/\/accounts\/[\w|-]*\/domains/, '{}').respond(201, {
```
nreal: PUT Domain respond HttpStatusCode.OK(200) 
https://github.com/MakingSense/Doppler_Relay/pull/830/files#diff-f71167727d6150eec939131d77d53d79R373, maybe is possible change the number 201 by 200 ?

PR #19, opened 2017-03-29
DR-966 Add view, controller, service and test for deleting a domain

mvillaseco:
Changes:
 - Add Test.
 - Add controller.
 - Add service.
 - Change html.

### DIFF
--- a/src/spec/page-objects/settings-page.js
+++ b/src/spec/page-objects/settings-page.js
@@ -8,6 +8,7 @@ class SettingsPage {
     this._domainInput = $('.domain--default-container input');
     this._domainInputButton = $('.domain--default-container button');
     this._domainsSetDefaultButtonList = element.all(by.css('.domain--default-action'));
+    this._domainDeleteButton = element.all(by.css('.domain--erase-button'));
     this._defaultDomain = $('.domain--default-container .default-domain');
     this._activateButtons = element.all(by.css('.domain--activate-action'));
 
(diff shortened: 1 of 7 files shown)


src/spec/page-objects/settings-page.js
```diff
@@ -8,6 +8,7 @@ class SettingsPage {
     this._domainInput = $('.domain--default-container input');
     this._domainInputButton = $('.domain--default-container button');
     this._domainsSetDefaultButtonList = element.all(by.css('.domain--default-action'));
+    this._domainDeleteButton = element.all(by.css('.domain--erase-button'));
```
andresmoschini: What do you think about to pluralize it? It could represent more than one, right?

src/spec/page-objects/settings-page.js
```diff
@@ -50,6 +51,14 @@ class SettingsPage {
   submitAddDomain(){
     return this._domainInputButton.click();
   }
+  countDeleteButtons(){
+    return this._domainDeleteButton.count();
+  }
+  clickDeleteDomain(){
```
andresmoschini: What do you think about `clickFirstDeleteDomainButtton`?

src/wwwroot/controllers/DomainManagerCtrl.js
```diff
@@ -40,14 +40,20 @@
...
         settings.createOrActiveDomain(domain.name)
         .then(function() {
           domain.disabled = false;
         });
       }
 
+      vm.deleteDomain = function(domain, index){
+        settings.deleteDomain(domain)
+        .then(function() {
+          vm.domains.splice(index, 1);
```
andresmoschini: What do you think about to receive a domain object (in place of domainName and index) and remove the object from `vm.domains` array? I think that it is safe because if there ar multiple deletions simultaneously the index could be wrong

PR #20, opened 2017-03-30
Dr-971 Disable domain

mvillaseco:
Changes:
 - Add Test.
 - Rename parameters inside services.
 - Add changes in view.
 - Add changes in controller.


src/wwwroot/locales/en-translation.js
```diff
@@ -214,6 +214,7 @@
   "edit_text": "Edit",
   "active_text": "Activate",
   "domain-manger_set_as_default": "Make it your default domain",
+  "domain-manger_disable": "Disable",
```
nreal: Typo domain-manager

src/spec/settings_spec.js
```diff
@@ -178,4 +178,26 @@ describe('Settings Page', () => {
...
+    // Arrange
+    beginAuthenticatedSession();
+    browser.addMockModule('descartableModule2', () => angular
+      .module('descartableModule2', ['ngMockE2E'])
+      .run($httpBackend => {
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/domains/).respond(200, {
+          "domains": [{name: "relay.com"}, {name: "fromdoppler.com" }, {name: "makingsense.com", disabled: true }, {name: "makingsense12.com" }],
+          "defaultDomain": "relay.com"
+        });
+        $httpBackend.whenPUT(/\/accounts\/[\w|-]*\/domains/, '{"disabled":true}').respond(200, {});
```
andresmoschini: I think that the URL should include the domain name. What do you think about to, as a rule, add `$` prefix and `^` sufix to regular expressions to ensure to not forgot something?

mvillaseco: I think it's a great idea. I will add it into the other tests too.

PR #21, opened 2017-03-31
DR-958 Rename defaultDomain to default in api response parsing

mvillaseco:
Changes:
 - Rename default domain response in controller.
 - Rename default domain in test mocking.


andresmoschini:
👍

PR #23, opened 2017-04-03
DR-975 Clear input after submit

mvillaseco:
Changes:
 - Add a new service inside utils for reseting an input.
 - Refactor registration Ctrl and add it to domain Ctrl.


src/wwwroot/services/utils.js
```diff
@@ -74,5 +75,19 @@
       });
     }
 
+    function resetInput(vm, form, defaultValues) {
```
andresmoschini: should it be named `resetInput` or `resetForm`?

andresmoschini:
❤️

PR #24, opened 2017-04-04
DR-958 Fix issue with duplicated elements in table

mvillaseco:
Changes:
 - Change ng-repeat.
 - Add function to apply property to each new element, set default or disabled.
 - Remove old animation.


nreal:
❤️

PR #39, opened 2017-04-28
DR-1026 - Add ":" in all fields of Activation form

nreal:
Hi team,
I added ":" in all name field of the registration form.

Could you Review?


andresmoschini:
Please, verify it because I have seen some places where `:` is set directly in the view, not in the literal.

nreal:
I verified and doesn't found in the view that a field have ":",  the translation files have this characters

PR #43, opened 2017-05-03
DR-1028 Add language manual behaviour

mvillaseco:
Changes:
 - Add new global variable to know when the language change.
 - Add autofill after change for the countries and industries array.
 - Small content change.


andresmoschini:
❤️

PR #44, opened 2017-05-04
Dr-1035 Add dkim spf status column

mvillaseco:
Changes:
 - Add new icons.
 - Add new column.
 - Add new logic for icons.
 - Add content.


andresmoschini:
👍 

Only try to be explicit about how we should prepare our dev environment to see the icons.

PR #46, opened 2017-05-05
DR-1036 Replace text for icons in domain manager actions

mvillaseco:
Changes:
 - Added the icons instead of the delete and domain settings text.
 - Add the icons in view.
 

PD: To be noticed that i don't remove the other actions because that will be done when i finish the following task http://jira.makingsense.com/browse/DR-1033 in order to be able to merge this PR and still having the change to set as default, disable, and activate domains.


andresmoschini:
At a glance, it looks fine, but my eyes are not enough ready to quick review these changes in HTML and CSS. 

Please review it yourself twice before merge.

PR #48, opened 2017-05-10
Dr-1033 add dropdown into domain status

mvillaseco:
Changes:
 - Add Drop down into view.
 - Add function to predefine each domain choice options.
 - Style dropdown to match the prototype.
 - Fix tests.
 - Fix responsive issue.

### DIFF
--- a/src/wwwroot/controllers/DomainManagerCtrl.js
+++ b/src/wwwroot/controllers/DomainManagerCtrl.js
@@ -78,11 +78,11 @@
       };
 
       function setDefaultDomain(domain) {
-        settings.setDefaultDomain(domain)
+        settings.setDefaultDomain(domain.name)
         .then(function() {
           recentlyUpdated(vm.defaultDomain);
-          vm.defaultDomain = domain;
-          recentlyUpdated(domain);
+          vm.defaultDomain = domain.name;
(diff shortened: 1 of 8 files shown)


src/wwwroot/controllers/DomainManagerCtrl.js
```diff
@@ -23,6 +23,7 @@
       var vm = this;
       vm.regexDomain = "(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{0,62}[a-zA-Z0-9]\\.)+[a-zA-Z]{2,63}$)";
       vm.activationPromise = activate();
+      vm.statusList = [{ name: $translate.instant('default_text'), id : 1}, {name: $translate.instant('enable_text'), id : 2}, {name: $translate.instant('disabled_text'), id : 3}];
```
andresmoschini: Will it be updated when user change language? (I know that by the moment we do not allow that, but anyway...

andresmoschini:
Seems to be fine.

src/wwwroot/controllers/DomainManagerCtrl.js
mvillaseco: Yes it currently does.

PR #49, opened 2017-05-11
[DR-1063] Retrieve countries and industries from API

pbarrios:
## Background

In the confirmation page, we need to populate the drop-downs components with the JSON arrays provided from the API. So we need to remove the current hardcoded values and replace them by an AJAX call.

## Changes done

- Replace hardcoded values by a service method.
- Add tests for the confirmation page.
- Fix acceptance tests.


src/wwwroot/controllers/ConfirmationCtrl.js
```diff
@@ -38,8 +36,17 @@
 
     function fillList() {
       var lang = $translate.use();
-      vm.industryList = INDUSTRIES.map(function(val){ return { code: val.code, name: val[lang] }; });
-      vm.countryList = COUNTRIES.map(function(val){ return { code: val.code, name: val[lang] }; });
+      signup.getIndustries().then(function(ret) {
+        vm.industryList = ret.data.map(function(val){
```
andresmoschini: Does it make sense to cache result? or we are taking advance of HTTP cache?

src/wwwroot/partials/signup/confirmationPage.html
```diff
@@ -18,24 +18,24 @@
       <div class="flex margin--bottom-inputs">
         <div class="wrapper-input width--full margin--right-inputs">
           <label class="label--regular" for="industry">{{'industry_text' | translate}}:</label>
-          <ui-select ng-model="selected.value" name="industry" title="{{'select_choose_industry' | translate}}" on-select="form.industry.$modelValue = $model" required>
+          <ui-select id="confirmation-select-industry"  ng-model="selected.value" name="industry" title="{{'select_choose_industry' | translate}}" on-select="form.industry.$modelValue = $model" required>
```
pbarrios: @mvillaseco @andresmoschini as discussed with some UI developers, we believe that is a good practice to add identifiers to the form fields. They also facilitate the tests development.
In order to avoid collisions, the Id is created as [controllerName]-[inputType]-[fieldName]

andresmoschini:
Great! 

My only concern is about if it is working fine when the server response is a little slow. I mean, before we already had the values in memory, now we are loading them on loading the page o changing the language. If the server takes time in return the response, it will work fine anyway?

Another very minor thing is the indentation. It seems that there are some _tabs_. If I am not wrong we are trying to use two spaces indentation in place of tabs.

src/wwwroot/partials/signup/confirmationPage.html
andresmoschini: Personally, I prefer a simpler id, like `field_industry`, I do not imagine a collision situation now, and when it happen we could analyze the situation and find the better solution.

If you insist on defining a convention, I prefer something simpler like `[controllerName]-[fieldName]` or better yet `[formName]-[fieldName]`.

**BTW You should update the _for_ value from the related label.**

src/spec/confirmation_spec.js
```diff
@@ -0,0 +1,54 @@
+describe('Confirmation Page', () => {
+  var ConfirmationPage = require('./page-objects/confirmation-page').ConfirmationPage;
+
+  beforeEach(() => {
+      browser.addMockModule('descartableModule', () => angular
+          .module('descartableModule', ['ngMockE2E'])
+          .run($httpBackend => {
+             $httpBackend.whenGET(/\/resources\/industries\.json/).respond(200, [{"code": "dplr138","en": "Tobacco","es": "Tabaco"}]);
+             $httpBackend.whenGET(/\/resources\/countries\.json/).respond(200, [{"code": "BV","en": "The Bolivian people","es": "Bolivia"}]);
```
mvillaseco: Weird country to be used, but it's ok! 👍

src/spec/confirmation_spec.js
```diff
@@ -0,0 +1,54 @@
...
+             });
+             $httpBackend.whenPOST(/\/user\/apikeys/).respond(201, { "api_key": "akbjhwltbtpw60givp7vka8qbsv0xf" });
+      }));
+  });
+
+  afterEach(() => {
+    browser.removeMockModule('descartableModule');
+  });
+
+  it('should load drop downs in different languages', () => {
```
mvillaseco: I think it should be  "dropdown" instead of drop down.

src/wwwroot/partials/signup/confirmationPage.html
andresmoschini: It seems that there is a way to generate the input id, see here: https://github.com/angular-ui/ui-select/blob/32829a39e8efe7f94de049c6b529c23fedfc6686/src/uiSelectDirective.js#L34

What do you think to use it to apply the id to the generated input in place of the generated div (or span)?

mvillaseco:
GREAT PR!

https://www.google.com.ar/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&cad=rja&uact=8&ved=0ahUKEwiC2rzihujTAhVKlJAKHb6hDE0QjRwIBw&url=https%3A%2F%2Fmemesuper.com%2Fcategories%2Fview%2Fd2c127f7e6f43f877ce0c57185d763b63a1dea70%2Fthats-nice-meme.html&psig=AFQjCNFK0HvOS7vfLdnae6PuY_SQO5y2-w&ust=1494599388151317

mvillaseco:
💯

PR #50, opened 2017-05-11
DR-1067 Improve copy to clipboard

martinkeimel:
# Background
For the latests browsers there's no need to show a popup in order to copy a text to the clipboard so we are now using clipboardjs library in order to acomplish that. We'll still keep the popup as a fallback for older browsers

# Changes done
- Add clipboardjs library
- Create directive to handle all the copy related stuff. It's in charge of checking if the browser suppports direct copy and decide if the popup needs to be shown or not
- Applied some styling improvements suggested by Mariana

# Pending to be done
I'm working on fixing the e2e tests but I created this PR so you can start taking a look at the changes


src/wwwroot/partials/settings/connection-settings.html
```diff
@@ -37,7 +37,7 @@
       <label>{{'connection-settings_api_key_title' | translate}}</label>
       <span class="api-key--container">
         <p>{{vm.apiKey}}</p>
-        <button class="button button--extra-small" ng-click="vm.showCopy()">{{'connection-settings_copy_api_key' | translate}}</button>
+        <copy-to-clipboard copy-target=".api-key--container p" copy-tooltip-message-key="connection-settings_copy_success" button-text-key="connection-settings_copy_api_key"></copy-to-clipboard>
```
andresmoschini: @martinkeimel is `copy-target ` a standard angular attribute? 

In that case, @pbarrios, maybe it is useful to generate the _id_ in the inputs generated by the select plugin.

andresmoschini: No no, I understood it wrong! sorry.

src/wwwroot/styles/modules/_button.scss
```diff
@@ -85,9 +85,9 @@ $input--error--color: #FFF;
 		vertical-align: middle;
 
 		&.button--extra-small{
-			padding: toem(4px);
+			padding: toem(5px);
 			min-width: 100px;
-			font-size: $font-size-small;
+			font-size: $font-size-tiny;
```
mvillaseco: Good use of Sass Variables 👍

src/wwwroot/styles/views/_settings.scss
```diff
@@ -24,6 +24,9 @@
           display: inline-block;
         }
       }
+      span.api-key--container p {
+        padding-right: 25px;
```
mvillaseco: Can we put this padding with 'em'? so when we need to reduce it in responsive mode, we can do it by changing only the font size of the body?

src/wwwroot/partials/settings/connection-settings.html
martinkeimel: No prob! Just in case, it's an input for the newly created directive

src/wwwroot/styles/views/_settings.scss
martinkeimel: Done!

mvillaseco:
💯

PR #51, opened 2017-05-11
DR-1035 Fix issue with arrow

mvillaseco:
Change:
 - Change how the library was setting the right arrow in the dropdowns inside settings.


andresmoschini:
👍

PR #53, opened 2017-05-12
DR-1073 Avoid rendering password field in the registration form

martinkeimel:
# Background
When moving across the registration form fields using the TAB key, the form ends in an invalid state. This happens because we have a hidden password field which is required.

# Changes
Avoid rendering password field


andresmoschini:
Good catch @martinkeimel !

andresmoschini:
:shipit:

PR #56, opened 2017-05-12
DR-1069 Prevent form to be submitted more than once

martinkeimel:
# Background
On activation page, if the user clicks on the confirmation button several times, it generates several API calls

# Changes done
- Add new `debounced-ng-click` directive in charge of preventing new submits if a promise is in progress
- Adjust `submitActivation` function so it returns a promise

# Pending to be done
If this approach is approved we should consider using this directive in every element that causes an API call. In those cases, we should adjust controller code so the involved functions returns a promise


andresmoschini:
❤️

PR #60, opened 2017-05-24
Dr 794 change password

mvillaseco:
Changes: 
 - Add Change password functionality
 - Add Ctrl
 - Add service
 - Add view (to be refactored)
 - Add style (to be refactored)

### DIFF
--- a/src/wwwroot/controllers/MyProfileCtrl.js
+++ b/src/wwwroot/controllers/MyProfileCtrl.js
@@ -46,7 +46,13 @@
         $timeout(function(){
           vm.changePasswordSuccess = false;
         }, 1500);
-
+      })
+      .catch(function(rejectionData){
+        if (rejectionData.data.errorCode == 2 && rejectionData.data.status == 401) {
+          vm.wrongOldPassword = true;
+        } else {
+          $rootScope.addError('action_updating_password', rejectionData.data.detail, rejectionData.data.title, rejectionData.data.status, rejectionData.data.errorCode);
+        }
       });
     }
   }
--- a/src/wwwroot/controllers/MyProfileCtrl.js
+++ b/src/wwwroot/controllers/MyProfileCtrl.js
@@ -3,9 +3,9 @@
 
   angular
     .module('dopplerRelay')
-    .controller('BasicInfoCtrl', BasicInfoCtrl);
+    .controller('MyProfileCtrl', MyProfileCtrl);
 
-  BasicInfoCtrl.$inject = [
+  MyProfileCtrl.$inject = [
     '$scope',
     '$location',
     '$rootScope',
@@ -13,10 +13,10 @@
     '$translate'
   ];
 
-  function BasicInfoCtrl($scope, $location, $rootScope, auth, $translate) {
+  function MyProfileCtrl($scope, $location, $rootScope, auth, $translate) {
     var vm = this;
     $rootScope.setSubmenues([
-      { text: 'submenu_basic_info', url: 'settings/basic-info', active: true },
+      { text: 'submenu_my_profile', url: 'settings/my-profile', active: true },
       { text: 'submenu_smtp', url: 'settings/connection-settings', active: false },
       { text: 'domains_text', url: 'settings/domain-manager', active: false }
     ]);
--- a/src/wwwroot/controllers/MyProfileCtrl.js
+++ b/src/wwwroot/controllers/MyProfileCtrl.js
@@ -10,10 +10,11 @@
     '$location',
     '$rootScope',
     'auth',
-    '$translate'
+    '$translate',
+    '$timeout'
   ];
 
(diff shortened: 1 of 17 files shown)


src/wwwroot/controllers/MyProfileCtrl.js
```diff
@@ -0,0 +1,53 @@
...
+    ]);
+    vm.changePassword = changePassword;
+    vm.updateValidation = updateValidation;
+
+    function updateValidation(newPass, confNewPass) {
+      if (!$scope.form.pass.$modelValue || !$scope.form.confPass.$modelValue) {
+        $scope.form.confPass.$setValidity('same', null);
+      } else if ($scope.form.pass.$modelValue != $scope.form.confPass.$modelValue) {
+        $scope.form.confPass.$setValidity('same', false);
+      } else {
```
pbarrios: If in all cases you're setting a value you can avoid this extra else:

```javascript
 function updateValidation(newPass, confNewPass) {
   $scope.form.confPass.$setValidity('same', true);
   if (!$scope.form.pass.$modelValue || !$scope.form.confPass.$modelValue) {
     $scope.form.confPass.$setValidity('same', null);
    } else if ($scope.form.pass.$modelValue != $scope.form.confPass.$modelValue) {
     $scope.form.confPass.$setValidity('same', false);
    }
  }
```

src/wwwroot/locales/es-translation.js
```diff
@@ -262,5 +262,11 @@
   "domain_manager_dkim_spf_status_th": "SPF & DKIM",
   "domain_manager_status_error_tooltip": "El DKIM no está configurado aún.",
   "domain_manager_status_warning_tooltip": "El SPF no está configurado aún.",
-  "select_choose_status": "Selecciona estado"
+  "select_choose_status": "Selecciona estado",
+  "submenu_my_profile": "Mi Perfil",
+  "my_profile_title": "Mi Perfil",
+  "my_profile_subtitle": "Password",
+  "my_profile_old_password":"Old Password",
```
pbarrios: Don't you have to translate these sentences to spanish?

src/wwwroot/services/auth.js
```diff
@@ -197,5 +198,25 @@
...
+        data: {
+          "username": getUserName(),
+          "old_password": old_pass,
+          "password": newPass
+        }
+      })
+      .then(function (response) {
+        return response;
+      })
+      .catch(function (reason) {
```
pbarrios: If the promise fails we're not doing anything. Is that OK? If so, why adding the catch function?

src/wwwroot/controllers/MyProfileCtrl.js
```diff
@@ -0,0 +1,53 @@
...
+
+    function changePassword(form) {
+      vm.submitted = true;
+      if (form.pass.$modelValue != form.confPass.$modelValue || !form.$valid) {
+        return;
+      }
+      auth.changePassword(form.oldPass.$modelValue, form.pass.$modelValue, $translate.use())
+      .then(function() {
+        vm.showChangePassContainer = false;
+        vm.changePasswordSuccess = true;
```
pbarrios: Is the change password always successful? Don't we have to validate the response from the server? Or to catch an exception thrown during the promise execution?

pbarrios:
Thanks @mvillaseco! Is it possible to add some tests to this amazing work you have done?

src/wwwroot/controllers/MyProfileCtrl.js
andresmoschini: I like @mvillaseco approach, is more like a _switch_, at least for me it is easier to read.

src/wwwroot/services/auth.js
```diff
@@ -197,5 +198,25 @@
...
+          'Content-Type': 'application/json'
+        },
+        data: {
+          "username": getUserName(),
+          "old_password": old_pass,
+          "password": newPass
+        }
+      })
+      .then(function (response) {
+        return response;
```
andresmoschini: This _then_ is redundant. Could you remove it?

src/wwwroot/styles/views/_settings.scss
```diff
@@ -199,6 +199,109 @@
...
+    border-top: $suit-border-default;
+  }
+  .item:last-of-type {
+    padding-bottom: 2.5em;
+    border-bottom: $suit-border-default;
+  }
+  .password.item{
+    overflow: hidden;
+    input{align-items: flex-end;}
+    transition: height 0.3s ease;ight: toem(245px);
```
andresmoschini: typo?

src/wwwroot/services/auth.js
mvillaseco: Yeah, i'll remove the catch, and if the promise fails the error handler will catch that error.

PR #63, opened 2017-05-29
DR-1074 - Dkim Configuration Page - Change style of subtitle

nreal:
Hi,
I change the style of subtitle text in SPF and Dkim Configuration.
Example:

Could you review?

### DIFF
--- a/src/wwwroot/partials/settings/dkim-configuration-help.html
+++ b/src/wwwroot/partials/settings/dkim-configuration-help.html
@@ -10,7 +10,7 @@
   <hr class="default" />
 </div>
 <div class="item section--wrapper flex">
-  <h3><strong>{{ 'dkim_configuration' | translate }}</strong></h3>
+  <h3 class="special">{{ 'dkim_configuration' | translate }}</h3>
 </div>
 <div class="item section--wrapper dkmin-config-code">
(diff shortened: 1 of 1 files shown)


src/wwwroot/partials/settings/dkim-configuration-help.html
```diff
@@ -10,9 +10,7 @@
   <hr class="default" />
 </div>
 <div class="item section--wrapper flex">
-  <h2>
-    {{ 'dkim_configuration' | translate }}
-  </h2>
+  <h3><strong>{{ 'dkim_configuration' | translate }}</strong></h3>
```
andresmoschini: Is _strong_ necessary? I mean, are there other _h3_ headers without strong? if not, why not adding bold style to _h3_? if there are, why? @mvillaseco  what do you think?

mvillaseco: Yeah, we should not add a bold into any h3 header, we have several others h3 in the app and no one have bold font weight, so, lets have a small meeting so we can work this out.

mvillaseco: After a research i found that in the _typography.scss file we only have h1 and h2 as bold, but below of those definitions there is a class named "special" that adds bold and a line-height into any of the other headers, i think you should use that class.

nreal: Done

PR #64, opened 2017-05-31
DR-1112 Fix submenues to be used in the future

mvillaseco:
TLDR: This PR changes the submenues making it like this:
 - My Profile, My Plan and Billing
 - SMTP & API and Domain Manager

and the only way to access between them it's from the upper right menu.
Changes:
 - Modified sub menues
 - Add menu into upper right menu.

WIP: I'm waiting for final Content.


mvillaseco:
The submenues that points into a non existent page, will redirect you to Reports, off course this it's just temporal because when we have those pages it will work OK.

andresmoschini:
👍

PR #70, opened 2017-06-09
Dr-1132 - Plan names

andresmoschini:
Based on our yesterday conversation with _Juampi_ and _Luciana_, we decided to avoid plan name translation. The name will be the same for English and Spanish.

Today we adopted the follow convention with @diegofox2: `PLAN-10K`, `PLAN-60K`, `PLAN-300K`, etc.

So, I have updated the default plan name and remove the translations.

Could you review?

@nreal it should not affect backend code, right?

cc: @mvillaseco


mvillaseco:
This looks great!

PD: Remind to update DB plans name to match this new standard (so the default plan will be selected).

andresmoschini:
> PD: Remind to update DB plans name to match this new standard (so the default plan will be selected).

Yep, @diegofox2 is taking care of that, see [DR-1129](http://jira.makingsense.com/browse/DR-1129)

nreal:
No, only pay attention when active a new user the system find DEFAULT_PLAN if change this name should be changed in the config.JSON key DefaultFreePlan in Application.API project

PR #73, opened 2017-06-13
DR-1156 - Plans cache

andresmoschini:
It is to avoid to request our API when the plan list is already loaded in client memory.

In the future we should avoid this kind of patches and make a goo use of HTTP cache, but I think that by the moment it is useful.

Could you review?


mvillaseco:
Nice!

PR #74, opened 2017-06-13
DR-1159 Add plan details into billing page with not founded plan behaviour

mvillaseco:
# Overview
 Shows the selected price with his price and also add a behaviour of select the first plan when the plan selected it's not available.

# Changes

-  Modify Billing controller. 
-  Modify html and css.
-  Add a load function.

### DIFF
--- a/src/wwwroot/controllers/BillingCtrl.js
+++ b/src/wwwroot/controllers/BillingCtrl.js
@@ -21,19 +21,23 @@
     $rootScope.setSubmenues([
       { text: 'submenu_my_profile', url: 'settings/my-profile', active: false },
     ]);
-    vm.activationPromise = activate();
     var queryParams = $location.search();
-    vm.planName = queryParams['plan'];
+    var planName = queryParams['plan'];
+    vm.activationPromise = activate();
+
     function activate() {
-      var planSelected;
+      if (!planName) {
+        return $location.path('/settings/my-plan');
+      }
+      vm.planName = planName;
       return settings.getPlansAvailable().then(function(response){
-        planSelected = response.data.items.find(function(obj){
-          return obj.name == vm.planName;
+        var planSelected = response.data.items.find(function(obj){
+          return obj.name == planName;
         });
         if (!planSelected) {
-          planSelected = response.data.items[0];
-          vm.planName = planSelected.name;
+          return $location.path('/settings/my-plan');
         }
+        vm.currentCurrency = planSelected.currency;
         vm.planPrice = planSelected.fee;
       });
     }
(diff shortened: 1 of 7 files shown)


src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -21,6 +21,22 @@
...
+    var queryParams = $location.search();
+    vm.planName = queryParams['plan'];
+    function activate() {
+      var planSelected;
+      return settings.getPlansAvailable().then(function(response){
+        planSelected = response.data.items.find(function(obj){
+          return obj.name == vm.planName;
+        });
+        if (!planSelected) {
+          planSelected = response.data.items[0];
```
andresmoschini: What do you think about to redirect to `my-plan` when the plan is not valid?

andresmoschini:
What do you think about adding some little test?

src/wwwroot/partials/settings/billing.html
```diff
@@ -5,6 +5,18 @@
...
   </div>
+  <div class="section--wrapper billing--plan-name-container">
+    <hr class="default"/>
+    <div>
+      <label class="uppercase">{{ 'plan_text' | translate }}</label>
+      <h4 class="special uppercase">{{vm.planName}}</h4>
+    </div>
+    <div>
+      <label class="uppercase">{{ 'price_text' | translate }}</label>
+      <h4 class="special">${{vm.planPrice}} {{ 'each_month' | translate }}</h4>
```
nreal: The symbol of `$` you can be changed? because this symbol you can get in of backend

andresmoschini: Good catch @nreal we should use currency field, because we will use the same in the email

mvillaseco: From backend i only got the country currency identifier, the icon $ will still be there. Is that ok?

andresmoschini: Currency values are [official currency names](https://en.wikipedia.org/wiki/ISO_4217), for example for USA Dollars, USD is the right name.

JP accepted to show that name in place of _dollars_ or _$_.

src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -21,6 +21,27 @@
...
+    vm.activationPromise = activate();
+
+    function activate() {
+      if (!planName) {
+        return $location.path('/settings/my-plan');
+      }
+      vm.planName = planName;
+      var planSelected;
+      return settings.getPlansAvailable().then(function(response){
+        planSelected = response.data.items.find(function(obj){
```
andresmoschini: What do you think about to declare the variable `planSelected` directly here?

PR #78, opened 2017-06-19
Dr 1153 - Billing page - Add Country combo

nreal:
Hi team,
I added country combo in the Billing page with the same countries of activation flow
Example:

Could you review?

### DIFF
--- a/src/spec/billing_spec.js
+++ b/src/spec/billing_spec.js
@@ -32,6 +32,7 @@ describe('Billing Page', () => {
               "name": "PLAN-60K" }
           ]
         });
+       $httpBackend.whenGET(/\/resources\/countries\.json/).respond(200, [{"code": "BV","en": "The Bolivian people","es": "Bolivia"}]); 
       }));
   }
 
@@ -39,22 +40,7 @@ describe('Billing Page', () => {
     // Arrange
     beginAuthenticatedSession();
     browser.get('/#/settings/billing?plan=PLAN-60K');
-    browser.addMockModule('descartableModule2', () => angular
-      // This code will be executed in the browser context,
-      // so it cannot access variables from outside its scope
-      .module('descartableModule2', ['ngMockE2E'])
-      .run($httpBackend => {
-        $httpBackend.whenGET(/\/plans/).respond(200, {
-          "items": [
-            { "currency": "USD",
-              "fee": 5.90,
-              "name": "PLAN-10K"},
-            { "currency": "USD",
-              "fee": 31.8,
-              "name": "PLAN-60K" }
-          ]
-        });
-      }));
+    setupSamplePlansResponse();
 
     var billingPage = new BillingPage();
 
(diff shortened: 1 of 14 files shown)


src/spec/billing_spec.js
```diff
@@ -32,29 +32,15 @@ describe('Billing Page', () => {
               "name": "PLAN-60K" }
           ]
         });
+       $httpBackend.whenGET(/\/resources\/countries\.json/).respond(200, [{"code": "BV","en": "The Bolivian people","es": "Bolivia"}]);
```
mvillaseco: "The Bolivian people" is this ok?

src/spec/page-objects/billing-page.js
```diff
@@ -7,6 +7,7 @@ class BillingPage {
     this._ccIconAmex = $('.icon-amex');
     this._ccIconVisa = $('.icon-visa');
     this._ccIconMaster = $('.icon-master');
+    this._countrySelect =  element(by.id('country'));
```
mvillaseco: It's not a must, but why did you use "element(by.id(" instead of " $('# "?

mvillaseco:
Great!

src/spec/billing_spec.js
nreal: No, https://github.com/MakingSense/Doppler_Relay/blob/develop/Relay.Application.Api/wwwroot/resources/countries.json#L172 I change the name of country, thanks

src/spec/page-objects/billing-page.js
nreal: I prefer `by.id('country')` because is more clear for me, but I verify with `instead of`

src/spec/billing_spec.js
nreal: Done

src/spec/page-objects/billing-page.js
```diff
@@ -27,5 +28,23 @@ class BillingPage {
   isCcIconAmexDisplayed() {
       return this._ccIconAmex.isDisplayed();
   }
+  getFirstCountryName() {
+    return this.getFirstSelectOptionText(this._countrySelect);
+  }
+  getFirstSelectOptionText(parentElem)
+  {
```
andresmoschini: In general, we are using the curly brace in the same line than function name

src/spec/page-objects/billing-page.js
```diff
@@ -27,5 +28,23 @@ class BillingPage {
   isCcIconAmexDisplayed() {
       return this._ccIconAmex.isDisplayed();
   }
+  getFirstCountryName() {
+    return this.getFirstSelectOptionText(this._countrySelect);
```
andresmoschini: Is it the first? or is it the selected one?

src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -121,6 +128,16 @@
...
+
+    function fillList() {
+      var lang = $translate.use();
+
+      countriesPromise.then(function(ret) {
+        vm.countryList = ret.data.map(function(val){
+          return { code: val.code, name: val[lang] };
+        })
+      });
+    }
```
andresmoschini: Maybe we could move all this common country data to a single place, it could avoid registration and deregistration complications

src/wwwroot/services/settings.js
```diff
@@ -137,5 +138,18 @@
 
       return plansCache;
     }
+
+    function getCountries() {
```
andresmoschini: This code should already be implemented in another place.

src/spec/page-objects/billing-page.js
andresmoschini: If `$('#` works fine, I prefer to keep that way, in order to have the same style (I think that, at the moment, we have more _css selector like_ ones than _by_ ones.

But, if it has any advantage, we can all switch to _by_ ones

src/spec/page-objects/billing-page.js
nreal: Is the first country name in the selector

andresmoschini: I understand now, _select_ is the kind of element hehe. Sorry.

src/wwwroot/services/settings.js
nreal: Yes in the `signup `service, do you think about injecting `signup `service on the `billing `controller in use it?

andresmoschini: mmmm, no, I am thinking about something more like creating a new _resources service_ and encapsulate there all of our weird code related to query API and language issues. 

What do you think?

nreal: yes, because injecting `sign up` service in the `billing `controller is not logic, I prefer that the code writing is clear, I will work about this

PR #80, opened 2017-06-23
DR-1175 Fix button text, API url typo and api values sent

mvillaseco:
# Overview
Fix several issues

# Changes
 - Fix text in buy button
 - Fix API url type
 - Fix Values sent to API.

### DIFF
--- a/src/wwwroot/controllers/BillingCtrl.js
+++ b/src/wwwroot/controllers/BillingCtrl.js
@@ -137,12 +137,12 @@
       vm.viewExpDate = form.expDate.$viewValue;
     }
     function submitBillingPayment() {
-      var agreement = {
+      var agreements = {
          plan : planName,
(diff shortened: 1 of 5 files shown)


src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -137,8 +137,12 @@
       vm.viewExpDate = form.expDate.$viewValue;
     }
     function submitBillingPayment() {
-      var paymentData = { plan : planName, creditCard: { number: vm.cc.number, securityCode: vm.secCode.number, expiration: vm.expDate} };
-      return settings.billingPayment(paymentData)
+      var agreements = {
```
nreal: for some reason this name is plural? because  think that should be singular

nreal:
👍

src/wwwroot/controllers/BillingCtrl.js
mvillaseco: Yes, it should be agreements because it's an object of several items.

andresmoschini:
:shipit:

PR #85, opened 2017-06-27
DR-1167 Add credit card validation with Luhn validation also

mvillaseco:
# Overview
 Add credit card validation and also Luhn algorithm.

# Changes
 - Add new method in utils service.
 - Add Validation error for this case.
 - Add tests.


src/wwwroot/services/utils.js
```diff
@@ -104,5 +105,29 @@
         });
     }
 
+    //This was taken from: https://gist.github.com/DiegoSalazar/4075533
+    function validateCreditCard(value) {
```
andresmoschini: Could you add some unit tests for it?

mvillaseco:
Is this good to go? :)

PR #87, opened 2017-07-03
Dr-1194 show plan values

mvillaseco:
# Overview
 Add plan information layout and some tests.

# Changes:
 - Add plan information layout.
 - Set service and controller.
 - Add Tests.

### DIFF
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -28,9 +28,18 @@
     vm.langUsed = $translate.use();
     vm.showPricingChart = showPricingChart;
     vm.pricingChartDisplayed = false;
+    vm.planInfoLoader = true;
 
     function activate() {
-      return settings.getPlansAvailable().then(function(response){
+      settings.getCurrentPlanInfo().then(function(response) {
+        vm.currentPlanPrice = response.data.fee;
+        vm.currentPlanEmailsAmount = response.data.includedDeliveries;
+        vm.currentPlanEmailPrice = response.data.extraDeliveryCost;
+      })
+      .finally(function () {
+        vm.planInfoLoader = false;
+      });
+      return settings.getPlansAvailable().then(function(response) {
         planItems = response.data.items;
         loadSlider();
         changePlan(defaultPlanName);
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -31,19 +31,21 @@
     vm.planInfoLoader = true;
 
     function activate() {
-      settings.getCurrentPlanInfo().then(function(response) {
+      var getCurrentPlanInfo = settings.getCurrentPlanInfo().then(function(response) {
         vm.currentPlanPrice = response.data.fee;
         vm.currentPlanEmailsAmount = response.data.includedDeliveries;
         vm.currentPlanEmailPrice = response.data.extraDeliveryCost;
+        vm.currency = response.data.currency;
       })
       .finally(function () {
         vm.planInfoLoader = false;
       });
-      return settings.getPlansAvailable().then(function(response) {
+      var getPlansAvailable = settings.getPlansAvailable().then(function(response) {
         planItems = response.data.items;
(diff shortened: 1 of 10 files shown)


src/wwwroot/locales/es-translation.js
```diff
@@ -309,5 +310,12 @@
...
   "billing_change_plan_button": "Seleccionar otro plan",
-  "billing_change_info_button": "Modificar información"
+  "billing_change_info_button": "Modificar información",
+  "my_plan_subtitle": "Encuentra aquí toda la información sobre tu plan y elige un nuevo plan si necesitas realizar una actualización.",
+  "amount_of_emails_text": "Cantidad de Emails",
+  "my_plan_cost_email": "Costo por email excedente",
+  "my_plan_consumption": "Mis consumos",
+  "my_plan_emails_sent": "Excedentes",
+  "my_plan_renewal_date": "Renovación de plan",
+  "change_plan_text": "Change your plan"
```
nreal: This text should be in spanish

src/wwwroot/services/settings.js
```diff
@@ -161,5 +163,19 @@
...
     }
 
+    function getCurrentPlanInfo() {
+      currentPlanInfoCache = currentPlanInfoCache || $http({
+        actionDescription: 'action_getting_current_plan',
+        method: 'GET',
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + auth.getAccountName() + '/agreements' + '/current'
+      }).catch(function(reason) {
+        currentPlanInfoCache = null;
+        console.log(reason);
```
nreal: I think that this console.log is not necessary

src/wwwroot/services/settings.js
```diff
@@ -161,5 +163,18 @@
       });
     }
 
+    function getCurrentPlanInfo() {
```
andresmoschini: I am not sure if it is a good idea to cache this result. What do you think @mvillaseco ?

mvillaseco: Mmm, i think that maybe if the user regrets the chosen plan, and come back to myplan selection, in that case that call will be made again. We should make this as fast as possible to the user.(also we have the other call of this page cached)

andresmoschini: mmm, I understand the scenario, but I am still not sure. For example, if we keep this cache, we should invalidate it when the user upgrades his account.

In my opinion, it could be a possible optimization in the future, but by the moment it looks as a premature optimization. In the same way, we could cache any request in the WebApp, and we are not doing that.

I think that the _List of Plans_ is a different scenario because the plans, does not change soooo much, and also it is a potentially expensive resource because it is a list of different plans.

Please, remove this cache, we could check how it works. But we could continue thinking in a way to reuse this logic and simplify cache invalidation, to use it when it will be necessary.

mvillaseco:
Is this good to go?

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -26,9 +26,21 @@
     var defaultPlanName = 'PLAN-60K';
     var planItems;
     vm.langUsed = $translate.use();
+    vm.showPricingChart = showPricingChart;
+    vm.pricingChartDisplayed = false;
+    vm.planInfoLoader = true;
 
     function activate() {
-      return settings.getPlansAvailable().then(function(response){
```
andresmoschini: The return value is still returning a promise that does not represent all activation work.

src/wwwroot/services/settings.js
```diff
@@ -161,5 +163,15 @@
...
+      return $http({
+        actionDescription: 'action_getting_current_plan',
+        method: 'GET',
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + auth.getAccountName() + '/agreements' + '/current'
+      }).catch(function(reason) {
```
andresmoschini: This catch is not necessary anymore

mvillaseco:
:shipit: ?

andresmoschini:
> :shipit: ?

I think so, it looks great and has nice tests!

PR #89, opened 2017-07-06
Dr-1203 show current plan status

mvillaseco:
# Overview
 Add current plan status from user.

# Changes
 - Add tests
 - Add new behaviour
 - Modify getReports service call to allow use "per_page" filter

# In progress
 - Hide upgrade functionality when user is not free trial
 - Set hidden menu ready to be ready when we finished this functionality

### DIFF
--- a/src/spec/billing_spec.js
+++ b/src/spec/billing_spec.js
@@ -24,8 +24,6 @@ describe('Billing Page', () => {
       .module('descartableModule4', ['ngMockE2E'])
       .run($httpBackend => {
         $httpBackend.whenGET(/\/accounts\/[\w|-]*\/agreements\/current/).respond(200, {
-          "items": [
-            {
               "planName": null,
               "paymentMethod": null,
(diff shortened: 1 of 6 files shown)


src/spec/billing_spec.js
```diff
@@ -460,4 +460,29 @@ describe('Billing Page', () => {
...
+    var billingPage = new BillingPage();
+    var dateConcat = '';
+    var date = new Date();
+    var month = ('0' + (date.getMonth() + 1 + 1)).slice(-2);
+    var year = date.getFullYear();
+    if (month > 12) {
+      month = '01';
+      year = year + 1;
+    }
+    dateConcat = dateConcat.concat(year,'-', month, '-', '01');
```
andresmoschini: This test, will test a different value each day, I understand that it is difficult to mock the time generator right now. So, it is fine for the moment. But if someone can think in an easy way to hard code the date, I think that it will be nicer an safer.

PR #90, opened 2017-07-06
[DR-1208] Add link to my plan section and free trial behaviour

mvillaseco:
# Overview
 Add link so the users can get into My Plan section now.

# Changes
 - Add link into submenu and settings menu.
 - Hide upgrade button for non free trial users.
 - Replace price and cost per emails for Free trial users.
 - Change transition for spinner, it was too slow.


nreal:
It's ok for me

PR #94, opened 2017-07-19
[Hot Fix] [DR-1158] Make API calls through HTTPS only

pbarrios:
## Background

This PR is intended to change the URL set up in the configuration files to make the calls through HTTPS in order to avoid the browser security errors


andresmoschini:
It is great! thanks @pbarrios

PR #97, opened 2017-07-21
[DR-1245] Remove USD in English

diegofox2:
New translation for "USD" so that each language can use the right version for USD.


andresmoschini:
Great @diegofox2! Two questions:

1. Did you search for all currencies shown in the WebApp?

2. Did you confirm with Juan Pablo that this is desired behavior?

diegofox2:
Yes, I talked with him and he told me that this is what he need.

PR #99, opened 2017-07-25
[DR-1253] - Add separator numbers by language

nreal:
Hi team,
I added the format separator number in all WebApp by language

This repository has a code to the use of this PR - angular-dynamic-locale code
https://github.com/lgalfaso/angular-dynamic-locale/blob/master/src/tmhDynamicLocale.js#L18

Example
https://user-images.githubusercontent.com/6796523/28466274-4c8b8206-6e02-11e7-9960-fbe93ae15d6c.png

Configuration i18n
Spanish
https://cdnjs.cloudflare.com/ajax/libs/angular-i18n/1.6.5/angular-locale_es-419.js
English
https://cdnjs.cloudflare.com/ajax/libs/angular-i18n/1.6.5/angular-locale_en-150.js

Could you review?

### DIFF
--- a/src/wwwroot/app.js
+++ b/src/wwwroot/app.js
@@ -23,7 +23,34 @@
     .filter('escapeURI', function(){
       return window.encodeURIComponent;
     })
-    .config(['$routeProvider', '$translateProvider', '$locationProvider', '$httpProvider', 'jwtInterceptorProvider', 'uiSelectConfig', 'tooltipsConfProvider', function ($routeProvider, $translateProvider, $locationProvider, $httpProvider, jwtInterceptorProvider, uiSelectConfig, tooltipsConfProvider) {
+    .config([
+      '$routeProvider', 
+      '$translateProvider', 
+      '$locationProvider', 
+      '$httpProvider', 
+      'jwtInterceptorProvider', 
+      'uiSelectConfig', 
+      'tooltipsConfProvider',
+      '$provide',
+      function (
+        $routeProvider,
+        $translateProvider,
+        $locationProvider,
+        $httpProvider,
+        jwtInterceptorProvider,
+        uiSelectConfig,
+        tooltipsConfProvider,
+        $provide) {
+
+      function makeStateful($delegate) {
+        $delegate.$stateful = true;
+        return $delegate;
+      }
+
+      $provide.decorator('dateFilter', ['$delegate', makeStateful]);
+      $provide.decorator('numberFilter', ['$delegate', makeStateful]);
+      $provide.decorator('currencyFilter', ['$delegate', makeStateful]);
+
 
       //  $locationProvider.html5Mode(true); //this apply HTML5MODE
       uiSelectConfig.theme = 'selectize';
@@ -133,7 +160,30 @@
 
     }]);
 
-  dopplerRelayModule.run(['$rootScope', 'auth', '$location', '$translate', 'jwtHelper', function ($rootScope, auth, $location, $translate, jwtHelper) {
+  dopplerRelayModule.run([
+    '$rootScope', 
+    'auth', 
+    '$location', 
+    '$translate', 
+    'jwtHelper', 
+    '$locale',
+    function (
+      $rootScope,
+      auth,
+      $location, 
+      $translate, 
+      jwtHelper,
+      $locale) {
+
+    function applyCulture() {
+      $locale.NUMBER_FORMATS.DECIMAL_SEP = $translate.instant("NUMBER_FORMATS.DECIMAL_SEP");
+      $locale.NUMBER_FORMATS.GROUP_SEP = $translate.instant("NUMBER_FORMATS.GROUP_SEP");
+    }
+
+    applyCulture();
+
+    $rootScope.$on('$translateChangeEnd', applyCulture);
+
     $rootScope.$on('$locationChangeStart', function () {
       var queryParams = $location.search();
 
--- a/src/wwwroot/app.js
+++ b/src/wwwroot/app.js
@@ -175,14 +175,14 @@
       jwtHelper,
       $locale) {
 
-    function applyCulture() {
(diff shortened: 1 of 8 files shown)


src/wwwroot/app.js
```diff
@@ -133,7 +160,30 @@
...
+    '$locale',
+    function (
+      $rootScope,
+      auth,
+      $location, 
+      $translate, 
+      jwtHelper,
+      $locale) {
+
+    function applyCulture() {
```
andresmoschini: What do you think about to rename it as `applyCultureFormats()`?

src/wwwroot/app.js
```diff
@@ -133,7 +160,30 @@
...
+      $locale.NUMBER_FORMATS.DECIMAL_SEP = $translate.instant("NUMBER_FORMATS.DECIMAL_SEP");
+      $locale.NUMBER_FORMATS.GROUP_SEP = $translate.instant("NUMBER_FORMATS.GROUP_SEP");
```
andresmoschini: What do you think to add a new level in our translate file to group all culture specific values? 

For example: 

```
$locale.NUMBER_FORMATS.GROUP_SEP = $translate.instant("CULTURE_FORMAT.NUMBER_FORMATS.GROUP_SEP");
```

andresmoschini: Also, we could take advantage of our translation variables convention and avoid `$translate.instant`, we could do something like:

```js
function applyCulture() {
  var lang = $translate.use();
  var cultureFormat = window['relay-translation-' + lang]['CULTURE_FORMAT'];
  angular.extend($locale, cultureFormat);
}
```

andresmoschini: @mvillaseco what do you think about this?

I know that we are cheating using our global variable, but, it is better than remembering to change it each time that we add a new value.

@diegofox2  you will add new formats for dates, what do you think?

src/wwwroot/app.js
```diff
@@ -133,7 +160,30 @@
...
+      auth,
+      $location, 
+      $translate, 
+      jwtHelper,
+      $locale) {
+
+    function applyCultureFormats() {
+      $locale.NUMBER_FORMATS.DECIMAL_SEP = $translate.instant("CULTURE_FORMATS.NUMBER_FORMATS.DECIMAL_SEP");
+      $locale.NUMBER_FORMATS.GROUP_SEP = $translate.instant("CULTURE_FORMATS.NUMBER_FORMATS.GROUP_SEP");
+    }
```
andresmoschini: > Also, we could take advantage of our translation variables convention and avoid `$translate.instant`, we could do something like:
> 
> ```js
> function applyCulture() {
>   var lang = $translate.use();
>   var cultureFormat = window['relay-translation-' + lang]['CULTURE_FORMAT'];
>   angular.merge($locale, cultureFormat);
> }
> ```
> 
> @mvillaseco what do you think about this?
> 
> I know that we are cheating using our global variable, but, it is better than remembering to change it each time that we add a new value.
> 
> @diegofox2  you will add new formats for dates, what do you think?
> 



I know that using the global variable is a smell, but it is an old smell hehe. Using _instant_ with generated keys, and making the merge manually is also a smell, right?

I applied my change in andresmoschini/relay-webapp@21c205b43ad6d5ede2810757467117b22808c3db it will be directly reusable for other formats like dates.

Could you take a look and merge in your branch if you like it?

andresmoschini:
👍

src/wwwroot/app.js
diegofox2: Yes, I talked with Nazareno and he showed me how is this approach, thank you!

src/wwwroot/partials/settings/my-plan.html
```diff
@@ -7,21 +7,21 @@
   </div>
   <div class="section--wrapper my-plan--info-container">
     <div class="item flex flex--wrap first">
-      <div class="flex one flex--wrap">
+      <div class="flex one flex--wrap current-plan-price">
```
diegofox2: Have you created a new rule style in a css file ?

diegofox2: mmm...ok. I think that we can improve this using [Data Attributes](https://www.w3schools.com/tags/att_global_data.asp) e.g. `<div class="flex one flex--wrap" data-test_identifier = "current-plan-price">`
I mean, I think that we have to avoid use css classes only for identify an element for UI test.  Html5 has this attribute so that we can store custom data like this.

What do you think @andresmoschini , @mvillaseco ?

andresmoschini: I understand that a class only for tests is a smell, but, in my opinion, a data attribute only for tests is not so different.

In my opinion, class attribute _is not for styles_ is for express semantics. Then, depending on the styles, that semantic could be represented with a special style or not. 

So, if the desired class express the semantics of the element, I mean, if this class could help a machine to understand the semantics of an element I think that it is a valid class. A machine could be a browser selecting the right style or Protractor trying to get an element with a particular meaning.

Of course, it could be better if we are consistent with the naming and how we represent those semantics using classes.

mvillaseco: If this is going to be a class for styling something then it's bad formatted, but if this it's only used as an identifier(i think this it's your case) OR a class inside another used for styling, then it's ok.

diegofox2: In my opinion, HTML5 provides data attribute for cases like this. However, we are using empty or fake classes only for identifying an element. If we really have a class with css styles it ok, but it is not that case.

nreal: Is  an interesting, for the moment I proceed to change the name of class by `price` and in the future, we will investigate as to create a locator in page object by `Data Attribute` because doesn't have an example in the WebApp

andresmoschini:
:shipit:

PR #103, opened 2017-07-27
[DR-1261] Change style for active status in every select box

mvillaseco:
# Overview
 Change stlye from all select box in the app to match the same style than the regular inputs.

# Changes:
 - Change in base.scss 
 - Change in dropdown.scss


andresmoschini:
👍

mvillaseco:
Default State:



Focus State:



Active State:

PR #105, opened 2017-07-31
[DR-1260] Change display when spinner is not showed in mobile

mvillaseco:
# OverView
 Remove horizontal scoll when the user goes to my plan in a mobile resolution. (this was caused by the spinner)

# Changes
 - Change display when mobile resolution is being used and the spinner should be hidden.

### DIFF
--- a/src/wwwroot/styles/views/_settings.scss
+++ b/src/wwwroot/styles/views/_settings.scss
@@ -588,7 +588,10 @@
   .my-plans--container{
     .my-plan--info-container{
       .item {
-        div {margin-top: 1em;}
+        div {          
+          margin-top: 1em;
+          .hide{display: none;}
+        }
         div:first-of-type { margin-top: 0; }
       }
        .item:first-of-type {


src/wwwroot/styles/views/_settings.scss
```diff
@@ -588,7 +588,10 @@
   .my-plans--container{
     .my-plan--info-container{
       .item {
-        div {margin-top: 1em;}
+        div {          
+          margin-top: 1em;
+          .hide{display: none;}
```
andresmoschini: There is not a general rule for this? I mean `.hide{display: none;}` at top level, because I imagine that we want to hide all elements marked with _hide_ class.

mvillaseco: Not in all hide cases we want to put a display none, this it's only for mobile versions(the width of the device do not allow us to use a nice transition and so then we can hide it without transition).

We should have a hide-transitioned class maybe with opacity or hide by max-height, but there is so much options to hide nicely an element.

andresmoschini: Thanks for your explanation @mvillaseco!

andresmoschini:
👍

PR #108, opened 2017-08-02
[DR-1261] Fix issue in domains with custom select box style

mvillaseco:
# Overview
It seems that when i made the change of styles for the select boxes override the custom select that it's in domains section that it's the only different from the rest of the application, so this is the fix.

# Changes:
 - Add a background-color defined to override the default.


andresmoschini:
We have talked in person with @mvillaseco to try to do not attach this custom style to a specific page.

andresmoschini:
Thanks @mvillaseco it looks great.

PR #109, opened 2017-08-03
[DR-1267] Add My plan link into submenu with false state

mvillaseco:
# Overview
 This is a fix for the meantime we decide what to do with that submenu.

# Change:
 - Add my plan link into submenu, so the user can go back without using the dropdown on the right top.

### DIFF
--- a/src/wwwroot/controllers/BillingCtrl.js
+++ b/src/wwwroot/controllers/BillingCtrl.js
@@ -34,6 +34,7 @@
     var vm = this;
     $rootScope.setSubmenues([
       { text: 'submenu_my_profile', url: 'settings/my-profile', active: false },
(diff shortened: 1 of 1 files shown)


andresmoschini:
👍 

Only by curiosity, do you have an idea about how we can reuse this code to avoid duplication and inconsistencies?

Maybe registering items in _rootScope_?

mvillaseco:
Maybe in rootscope we could have 2 possible submenus(like we have now), 2 links for one(Smtp&Apikey, Domain Manager) and 2 links for  the other(View Profile, My plan). And analize depending on the url wich submenu to use.

andresmoschini:
Thanks! we should take it into account to refactor as soon as possible.

PR #119, opened 2017-08-17
[DR-1313] Fix issue with price not adding ips values

mvillaseco:
# Overview
 Fix issue with value in billing page.

# Changes:
 - Add ip count and cost by ip values into the final price.


andresmoschini:
👍

PR #122, opened 2017-09-15
[DR-1341] Add daily rate limit

mvillaseco:
# Overview:
 Add limits by day into Reports page for users that have this limit.

# Changes:
 - Add limit into view, controller and service.
 - Manage locals.
 - Add test.
 - Responsive Handle

### DIFF
--- a/src/wwwroot/controllers/ReportsCtrl.js
+++ b/src/wwwroot/controllers/ReportsCtrl.js
@@ -29,7 +29,7 @@
     $scope.noLimits = true;
 
     function loadRateLimits() {
-      reports.getStatusPlanLimits()
+      return reports.getStatusPlanLimits()
       .then(function (result) {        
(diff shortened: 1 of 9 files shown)


src/wwwroot/controllers/ReportsCtrl.js
```diff
@@ -18,14 +18,30 @@
...
     $scope.canSearch = true;
     $scope.searchInProgress = false;
     $scope.curSection = $location.path().substring(1);
     $scope.deliveriesSummary = {};
     $scope.eventsSummary = {};
-
+    $scope.noLimits = true;
+
+    function loadRateLimits() {
+      reports.getStatusPlanLimits()
```
andresmoschini: You can prepend this line with return in order to allow _a potential client_ to consume the promise.

PR #124, opened 2017-09-19
[DR-1283] Add new informational header

mvillaseco:
# Overview
 Add new trial informational header letting the user know when it's in a trial account, about to end and in an ended account.

# Changes
 - Add free trial date to Rootscope so we can use it again in Reports Limits.
 - Add behaviour to show the correct text
 - Responsive ready.
 - Service ready to show the header as soon as we give to the webapp a freeTrialEndDate without making any changes.

### DIFF
--- a/src/wwwroot/controllers/HeaderCtrl.js
+++ b/src/wwwroot/controllers/HeaderCtrl.js
@@ -26,40 +26,6 @@
     };
     $scope.getSubmenues = $rootScope.getSubmenues;
     $scope.isSubmenuVisible = $rootScope.isSubmenuVisible;
-
-    UpdateTrialHeader();
-    
-    function UpdateTrialHeader() {
-      $scope.isFreeTrialEnded = false;
-      $scope.isFreeTrialAlmostEnded = false;
-      $scope.isFreeTrial = false;
-      return $rootScope.freeTrialEndDate().then(function(freeTrialEndDate) {
-        var todayDate = moment().toDate();
-        var daysLeft = moment(freeTrialEndDate).diff(todayDate, "days");
-        $scope.isFreeTrial = freeTrialEndDate ? true : false;
-        $scope.trialDaysLeft = daysLeft;
-        if (freeTrialEndDate <= todayDate) {
-          $scope.isFreeTrialEnded = true;
-          $scope.isFreeTrialAlmostEnded = false;
-          $scope.showFreeTrialHeader = false;
-        }
-        if (daysLeft <= 10 && daysLeft >= 0 && freeTrialEndDate > todayDate) {
-          $scope.isFreeTrialAlmostEnded = true;
-          $scope.isFreeTrialEnded = false;
-          $scope.showFreeTrialHeader = false;
-        }
-
-        if (freeTrialEndDate && !$scope.isFreeTrialEnded && !$scope.isFreeTrialAlmostEnded) {
-          $scope.showFreeTrialHeader = true;
-          $scope.isFreeTrialEnded = false;
-          $scope.isFreeTrialAlmostEnded = false;
-        }
-      });
-    }
-    
-    
-
-    
     $scope.initialsAvatar = function () {
       var fullName = $rootScope.getFullName();
       if (!fullName) {
--- a/src/wwwroot/controllers/HeaderCtrl.js
+++ b/src/wwwroot/controllers/HeaderCtrl.js
@@ -26,6 +26,40 @@
     };
     $scope.getSubmenues = $rootScope.getSubmenues;
     $scope.isSubmenuVisible = $rootScope.isSubmenuVisible;
+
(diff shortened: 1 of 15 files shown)


src/wwwroot/controllers/HeaderCtrl.js
```diff
@@ -26,6 +26,40 @@
...
+        var todayDate = moment().toDate();
+        var daysLeft = moment(freeTrialEndDate).diff(todayDate, "days");
+        $scope.isFreeTrial = freeTrialEndDate ? true : false;
+        $scope.trialDaysLeft = daysLeft;
+        if (freeTrialEndDate <= todayDate) {
+          $scope.isFreeTrialEnded = true;
+          $scope.isFreeTrialAlmostEnded = false;
+          $scope.showFreeTrialHeader = false;
+        }
+        if (daysLeft <= 10 && daysLeft >= 0 && freeTrialEndDate > todayDate) {
```
andresmoschini: Is it possible that `freeTrialEndDate > todayDate` evaluates as _true_ when `daysLeft <= 10 && daysLeft >= 0`? In what scenario does it happen?

src/wwwroot/controllers/HeaderCtrl.js
```diff
@@ -26,6 +26,40 @@
...
+    UpdateTrialHeader();
+    
+    function UpdateTrialHeader() {
+      $scope.isFreeTrialEnded = false;
+      $scope.isFreeTrialAlmostEnded = false;
+      $scope.isFreeTrial = false;
+      return $rootScope.freeTrialEndDate().then(function(freeTrialEndDate) {
+        var todayDate = moment().toDate();
+        var daysLeft = moment(freeTrialEndDate).diff(todayDate, "days");
+        $scope.isFreeTrial = freeTrialEndDate ? true : false;
```
andresmoschini: it could be simplified as `$scope.isFreeTrial = !!freeTrialEndDate;`

src/wwwroot/services/auth.js
```diff
@@ -214,5 +216,21 @@
...
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + getAccountName()  + '/status/limits'
+      }).then(function (response) {
+        if (!response.data.freeTrialEndDate){
+          return null;
+        }
+        return moment(response.data.freeTrialEndDate).toDate();
+      })
+      .catch(function (reason) {
+        return $q.reject(reason);
+      });
```
andresmoschini: This catch is not necessary, right?

src/wwwroot/services/auth.js
```diff
@@ -214,5 +216,21 @@
...
+    function getFreeTrialEndDate() {
+      return $http({
+        actionDescription: 'Gathering Free Trial end date',
+        method: 'GET',
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + getAccountName()  + '/status/limits'
+      }).then(function (response) {
+        if (!response.data.freeTrialEndDate){
+          return null;
+        }
+        return moment(response.data.freeTrialEndDate).toDate();
```
andresmoschini: What do you think about to store the complete response? in that way we could use the same request and object for daily limit box in reports page.

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -17,6 +17,8 @@
   ];
 
   function MainCtrl($rootScope, $scope, $window, $location, auth, $log, ModalService, $route) {
+    $rootScope.freeTrialEndDate = auth.getFreeTrialEndDate;
```
andresmoschini: What do you think about renaming `$rootScope.freeTrialEndDate` as `$rootScope.getFreeTrialEndDate`? I feel that without the verb it is not clear that it is a method, it looks like an accessor.

src/wwwroot/controllers/HeaderCtrl.js
```diff
@@ -26,6 +26,40 @@
     };
     $scope.getSubmenues = $rootScope.getSubmenues;
     $scope.isSubmenuVisible = $rootScope.isSubmenuVisible;
+
+    UpdateTrialHeader();
```
andresmoschini: Is it executed on each navigation?

I think that it is fine by the moment, but probably we will change it soon to be updated periodically.

src/wwwroot/services/auth.js
```diff
@@ -214,5 +216,15 @@
...
     }
+
+    function getLimitsByAccount() {
+      return $http({
+        actionDescription: 'Gathering Free Trial end date',
+        method: 'GET',
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + getAccountName()  + '/status/limits'
+      }).then(function (response) {
+        return response;
+      });
```
andresmoschini: This _then_ is not necessary.

andresmoschini:
👍

PR #125, opened 2017-09-21
Dr-1283 Add popup trial ended

mvillaseco:
# Overview
 Add new popup to show the user that the free trial has expired.

# Changes
 - Add new modal
 - Add content

# In progress
 - Tests

### DIFF
--- a/src/wwwroot/controllers/MainCtrl.js
+++ b/src/wwwroot/controllers/MainCtrl.js
@@ -18,7 +18,6 @@
   ];
 
   function MainCtrl($rootScope, $scope, $window, $location, auth, $log, ModalService, $route, $interval) {
-    $rootScope.getFreeTrialEndDate = auth.getFreeTrialEndDate;
     
     $rootScope.getLoggedUserEmail = function () {
       return auth.getUserName();
@@ -28,13 +27,13 @@
       return 1;
     };
      
-    $rootScope.getFreeTrialNotificationFromStorage = auth.getFreeTrialNotificationFromStorage();
+    var freeTrialNotification = auth.getFreeTrialNotificationFromStorage();
     var modalOpened = false;
     
     var loaderFreeTrial = function () {
       auth.getLimitsByAccount().then(function(limits){
         if (limits.freeTrialEndDate) {
-          UpdateTrialHeader(limits.freeTrialEndDate);
+          UpdateTrialHeader(moment(limits.freeTrialEndDate).toDate());
         }
       });
     }
@@ -48,7 +47,6 @@
       $rootScope.freeTrialStatus = null;
       return;
     }
-
     var todayDate = moment().toDate();
     var daysLeft = moment(freeTrialEndDate).diff(todayDate, "days");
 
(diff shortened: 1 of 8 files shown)


src/wwwroot/controllers/MainCtrl.js
```diff
@@ -63,7 +65,22 @@
...
+          controller: 'GeneralTemplateCtrl',
+          controllerAs: 'vm',
+          inputs: {
+            title: "free_trial_ended_popup_title",
+            mainText: "free_trial_ended_popup_subtitle",
+            buttonText: "free_trial_ended_popup_button_text",
+            action: freeTrialNotificated
+          }
+        }).then(function (modal) {
+          modal.close.then(toggleModalStatus);
```
andresmoschini: Is it not possible to define `modalOpened` as local variable in place of a `$rootScope`attribute.

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -63,7 +65,22 @@
         isFreeTrialAlmostEnded : false,
         isFreeTrialEndToday : false
       }
-      
+      if(!auth.getFreeTrialNotificationFromStorage() && !$rootScope.modalOpened) {
```
andresmoschini: You should compare stored date with current date in order to show again the modal if the trial was extended and expired again.

src/wwwroot/services/auth.js
```diff
@@ -223,6 +225,14 @@
...
         url: RELAY_CONFIG.baseUrl + '/accounts/' + getAccountName()  + '/status/limits'
       });
+    }   
+
+    function addFreeTrialNotificationToStorage(date) {
+      $window.localStorage.setItem('freeTrialNotificationOn', date);
+    }
+
+    function getFreeTrialNotificationFromStorage() {
+      return $window.localStorage.getItem('freeTrialNotificationOn');
```
andresmoschini: If I am not wrong, accessing to local storage has a cost, so, what do you think about read this value when the application is initialized and then use that value in place of read from local storage?

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -28,8 +28,11 @@
       return 1;
     };
      
+    $rootScope.getFreeTrialNotificationFromStorage = auth.getFreeTrialNotificationFromStorage();
```
andresmoschini: Maybe `getFreeTrialNotificationFromStorage` the right name for the variable could be `freeTrialNotificationDate`.

Also, maybe you could store it in a local variable in place of `rootScope`

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -63,7 +66,30 @@
         isFreeTrialAlmostEnded : false,
         isFreeTrialEndToday : false
       }
-      
+      console.log(moment($rootScope.getFreeTrialNotificationFromStorage).toDate());
+      console.log( moment(freeTrialEndDate).toDate());
+      console.log(moment($rootScope.getFreeTrialNotificationFromStorage) <= moment(freeTrialEndDate));
+      if (!modalOpened) {        
+        if(!$rootScope.getFreeTrialNotificationFromStorage || moment($rootScope.getFreeTrialNotificationFromStorage).toDate() <= moment(freeTrialEndDate).toDate()) {
```
andresmoschini: Is it necessary all of this transformation? in that case is it possible to do it in `getFreeTrialNotificationFromStorage` in order to do it less times?

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -81,6 +107,13 @@
         isFreeTrialEndToday : true
       }
     }
+
+    function freeTrialNotificated () {
+      auth.addFreeTrialNotificationToStorage(todayDate);
+      $rootScope.getFreeTrialNotificationFromStorage = auth.getFreeTrialNotificationFromStorage();
```
andresmoschini: Is it necessary to call `auth.getFreeTrialNotificationFromStorage()?, what do you think about to directly store `todayDate` in the local variable?

PR #128, opened 2017-09-22
[DR-1283] Fix class validations to add class for correct footer alignment

mvillaseco:
# Overview
 Fix issue with footer(scroll was being shown when it shouldn't).


andresmoschini:
I do not feel comfortable with that class notation, but if you tested it I am fine.

PR #130, opened 2017-09-26
[DR-1370] - Fix Free header upgrade is not displayed

nreal:
Hi,
The header is not displayed because of doesn't have `data ` object

Could you review?

### DIFF
--- a/src/wwwroot/controllers/MainCtrl.js
+++ b/src/wwwroot/controllers/MainCtrl.js
@@ -33,8 +33,8 @@
     var loaderFreeTrial = function () {
       if (auth.getUserName() != null) {
         auth.getLimitsByAccount().then(function(limits) {
-          if (limits.freeTrialEndDate) {
-            UpdateTrialHeader(moment(limits.freeTrialEndDate).toDate());
+          if (limits.data.freeTrialEndDate) {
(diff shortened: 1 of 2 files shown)


src/wwwroot/controllers/MainCtrl.js
```diff
@@ -33,8 +33,8 @@
     var loaderFreeTrial = function () {
       if (auth.getUserName() != null) {
         auth.getLimitsByAccount().then(function(limits) {
-          if (limits.freeTrialEndDate) {
-            UpdateTrialHeader(moment(limits.freeTrialEndDate).toDate());
+          if (limits.data.freeTrialEndDate) {
+            UpdateTrialHeader(moment(limits.data.freeTrialEndDate).toDate());
```
andresmoschini: 👍 

By the way, could you rename _freeTrialEndDate_ as _endDate_ or _accountEndDate_?

andresmoschini: Maybe, the responsibility of _unwrapping_ limits data is part of `auth.getLimitsByAccount()` function in place of `loaderFreeTrial()`

nreal: as `endDate` I'll rename

PR #133, opened 2017-09-28
[DR-1371] Remove trial period banner on logout

andresmoschini:
Hi @mvillaseco, @nreal and @alegarciamina,

As @nreal found yesterday, we were not cleaning free trial header after logout:

This code fixes that and also removes the header if the user upgrades the account. 

I have also applied a little code refactoring.

Could you kindly review?

### DIFF
--- a/src/wwwroot/services/auth.js
+++ b/src/wwwroot/services/auth.js
@@ -220,10 +220,17 @@
     }
 
     function getLimitsByAccount() {
+      var accountName = getAccountName();
+
+      if (!accountName) {
+        // limits have no sense in this scenario
+        return $q.when({});
+      }
+
       return $http({
         actionDescription: 'Gathering Free Trial end date',
         method: 'GET',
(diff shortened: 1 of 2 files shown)


src/wwwroot/services/auth.js
```diff
@@ -220,15 +220,51 @@
...
+      }
+    }
+
+    function mapDate(responseDate)
+    {
+      if (!responseDate) {
+        return null;
+      }
+
+      // TODO: parse date without using moment
```
nreal: Why is not using moment to parse a date?

andresmoschini: At the moment we are using moment to parse date and then converting it to a plain old JavaScript date: 
```
string => moment => date
```

I think that it is better to parse directly as date:
```
string => date
```

Or directly deal with moment dates:
```
string => moment
```

Does it make sense?

andresmoschini:
I am merging because I am creating a dependent PR, feel free to continue reviewing it if you want.

PR #136, opened 2017-10-12
[DR-1395] Issues on recurrent loading of limits

andresmoschini:
Hi @alegarciamina, there were some issues related to limits loading, I tried to fix one by commit, could you review?

I hope to deploy it to PROD environment tomorrow.

### DIFF
--- a/src/wwwroot/services/auth.js
+++ b/src/wwwroot/services/auth.js
@@ -228,7 +228,8 @@
       }
 
       return $http({
-        actionDescription: 'Gathering Free Trial end date',
+        actionDescription: 'Gathering account limits',
+        avoidStandarErrorHandling: true,
         method: 'GET',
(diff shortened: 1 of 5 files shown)


src/wwwroot/services/auth.js
```diff
@@ -262,20 +261,19 @@
...
     }
 
     function addFreeTrialNotificationToStorage(date) {
-      $window.localStorage.setItem('freeTrialNotificationOn', date);
+      $window.localStorage.setItem('freeTrialNotificationOn', date.toISOString());
     }
 
     function getFreeTrialNotificationFromStorage() {
       var storedData = $window.localStorage.getItem('freeTrialNotificationOn');
       if (!storedData) {
```
alegarciamina: We could remove this validation as is already covered by mapDate.

alegarciamina: We could remove this validation as is already covered by mapDate.

PR #141, opened 2017-10-25
[DR-1408] Error dialog callback

andresmoschini:
Hi team (and mainly @mvillaseco),

Before to fix the root cause of [DR-1374](http://jira.makingsense.com/browse/DR-1374), I wanted to fix a UX issue (in my humble opinion) with our error dialogs.

Error dialog were always reloading the page on closing, but sometime we could want to keep the page with the error, for example when there are useful information there, or sometimes it generates an cycling error.

So, I have remove _callback_ execution when the user closes the dialog with `x` icon.

I have also applied some minor code refactoring.

I am still testing but I am interested in know your opinion before spend so much time.

### DIFF
--- a/src/wwwroot/controllers/MainCtrl.js
+++ b/src/wwwroot/controllers/MainCtrl.js
@@ -115,42 +115,40 @@
     $rootScope.getFullName = auth.getFullName;
     $rootScope.getAccountId = auth.getAccountId;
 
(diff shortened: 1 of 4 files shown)


mvillaseco:
Hi, 
Yes i agree with you on this!

andresmoschini:
Closing in favor of #142

PR #142, opened 2017-10-25
[DR-1374] Error dealing with AJAX errors

andresmoschini:
Hi team, 

This PR fixes many issues on dealing with error responses, two of them related to authorization errors that where not redirecting to login.

It also includes a change in error dialog behavior: now we are executing the callback (refresh or redirect to login) only if the user clicks in the dialog button.

Could you review?


andresmoschini:
1xf6di

PR #145, opened 2017-10-26
[DR-1389] Fix close button from modals

mvillaseco:
# Overview
Fix missaligned close button from modals

# Changes:
 - Added proper class to align the button


andresmoschini:
👍

PR #148, opened 2017-10-30
[DR-1413] Use PUT agreements/current for upgrades

alegarciamina:
Hi team,
I've replaced `POST agreements` with `PUT agreements/current` for upgrades.

Could you review?


andresmoschini:
:shipit:

PR #152, opened 2017-10-31
[DR-1096] Persist preferred language on local storage

alegarciamina:


src/wwwroot/controllers/MainCtrl.js
```diff
@@ -14,11 +14,16 @@
...
+    '$translate',
+    'utils'
   ];
 
-  function MainCtrl($rootScope, $scope, $window, $location, auth, $log, ModalService, $route, $interval) {
-    
+  function MainCtrl($rootScope, $scope, $window, $location, auth, $log, ModalService, $route, $interval, $translate, utils) {
+     
+    var key = utils.getPreferredLanguage();
+    $translate.use(key);
```
andresmoschini: Is this code executed only one time on page loading or on each navigation?

andresmoschini: What happen when _key_ is null? should we deal with that scenario?

alegarciamina: It's executed on page loading only, just once.

alegarciamina: If key is null is just ignored and use English as default language.

PR #153, opened 2017-10-31
[DR-1421] Prevent multiple clicks on upgrade button

alegarciamina:
Disable submit button while processing the upgrade request.

### DIFF
--- a/src/wwwroot/controllers/BillingCtrl.js
+++ b/src/wwwroot/controllers/BillingCtrl.js
@@ -96,6 +96,7 @@
       allowInvalidValue: true, //allows us to watch the value
       clearOnBlur: false
     };
+    vm.processingPayment = false;
 
     function redirectToPlanSelection() {
       $location.path('/settings/my-plan');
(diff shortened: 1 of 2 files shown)


src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -176,7 +178,7 @@
         downgrade();
         return;
       }
-
+
```
andresmoschini: Very minor.  Remove this change.

src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -153,6 +154,7 @@
     }
 
     var onExpectedError = function (rejectionData) {
+      vm.processingPayment = false;
```
andresmoschini: What do you think about using _finally_ (or something like that)

andresmoschini:
:+1:

src/wwwroot/controllers/BillingCtrl.js
alegarciamina: User is redirected to My Plan on success, so it's not strictly necessary but finally looks cleaner and self-descriptive. I'll change it.

PR #155, opened 2017-11-03
[DR-1412] Add scheduled message in my plan

mvillaseco:
# Overview
 Show to the user a scheduled warning message

# Changes:
 - Add message
 - Mobile ready

### DIFF
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -44,10 +44,10 @@
         vm.currency = response.data.currency;
         vm.isFreeTrial = response.data.fee && response.data.includedDeliveries ? false : true;
         vm.currentIpsPlanCount = response.data.ips_count || 0;
-        vm.allowChangePlan = !response.data.endDate;
+        vm.hasScheduledPlan = !response.data.endDate;
         vm.hideDragMe = !vm.isFreeTrial;
         defaultPlanDeliveries = !vm.isFreeTrial ? response.data.includedDeliveries.toString() : defaultPlanDeliveries; 
-        if (!vm.allowChangePlan) {
+        if (!vm.hasScheduledPlan) {
           settings.getNextPlan().then(function(response){
             vm.nextPlan = response.data.includedDeliveries;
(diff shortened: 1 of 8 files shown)


src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -47,6 +47,11 @@
         vm.allowChangePlan = !response.data.endDate;
         vm.hideDragMe = !vm.isFreeTrial;
         defaultPlanDeliveries = !vm.isFreeTrial ? response.data.includedDeliveries.toString() : defaultPlanDeliveries; 
+        if (!vm.allowChangePlan) {
```
andresmoschini: I would rename it as _hasScheduledPlan_

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -47,6 +47,11 @@
...
+          settings.getNextPlan().then(function(response){
+            vm.nextPlan = response.data.includedDeliveries;
```
andresmoschini: It is not the name of the plan, what happen if it is PRO or something like that?

src/wwwroot/services/settings.js
```diff
@@ -181,6 +182,13 @@
         url: url
       });
     }
+    function getNextPlan() {
+      return $http({
+        actionDescription: 'action_getting_current_plan',
+        method: 'GET',
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + auth.getAccountName() + '/agreements' + '/next'
```
andresmoschini: I think that ` + '/agreements/next'` could be better if we search in the code-

src/wwwroot/controllers/PlanCtrl.js
mvillaseco: Yes, we are going to show only the amount of deliveries allowed for the plan the user select, we are currently not showing the name of the current plan.

andresmoschini: In that case maybe _nextPlan_ is not the best name

PR #162, opened 2017-11-09
[DR-1435] Send navigation events to Google Analytics

andresmoschini:
It sends navigation events to Google Analytics, only if `ga` object exists (so, it will only work in PROD).

Could you kindly review?


mvillaseco:
💯

PR #167, opened 2017-11-16
Dr-1457 Refactor succeed

mvillaseco:
# Overview
 Add recaptcha to succeed registration and small refactors.

# Changes:
 - Add recaptcha
 - Fixes in responsive.
 - Small fix in registration page

# Important:
 To Merge this i have to finish DR-1451 first


src/wwwroot/controllers/SucceedRegistrationCtrl.js
```diff
@@ -6,16 +6,49 @@
...
+            vm.widgetId = widgetId;
+        };
+        
+        function captchaExpiration () {
+            vcRecaptchaService.reload(vm.widgetId);
+            vm.response = null;
+        };
+
+        function resendEmail() {
+            vcRecaptchaService.reload(vm.widgetId);
```
andresmoschini: It looks weird, Is it necessary just before use captcha? 

If it is necessary, maybe it is a better idea to _reload_ during page initialization in place of doing after the click.

src/wwwroot/controllers/SucceedRegistrationCtrl.js
```diff
@@ -6,16 +6,49 @@
...
+            vcRecaptchaService.execute(vm.widgetId);
+        }
 
-        function getCurrentLanguage() {
-            return utils.getPreferredLanguage();
+        function setResponse (response) {
+            auth.forgotPassword(email, $translate.use(), response)
+            .then(function (result) {
+                 vm.resendFinished = true;
```
andresmoschini: If I understand it well, this line hides the "resend button".

In that case, I think that we should apply it just before call backend, for example:

```js
function setResponse (response) {
    vm.resendFinished = true;
    auth
        .forgotPassword(email, $translate.use(), response)
        .catch(function() {
            captchaExpiration();
            vm.resendFinished = false;
        });
}
```

Another option could be show a spinner, but I think that it is not necessary, the user does not need to know exactly when the email have been sent, at least not at millisecond.

Maybe we could also rename `resendFinished` as `resent`

mvillaseco:
all done in: https://github.com/DopplerRelay/relay-webapp/pull/168

PR #168, opened 2017-11-17
Dr-1461 Add recaptcha forgot password

mvillaseco:
# Overview
 Add invisible recaptcha to forgot password and refactor controller.
 

# Changes
 - Add invisible Recaptcha.
 - Remove $scope from controller.
 - Css Fixes


src/wwwroot/controllers/LoginCtrl.js
```diff
@@ -5,65 +5,95 @@
...
         }
       });
     };
 
-    $scope.removeSucceed = function () {
-      if ($scope.forgotSuccessful) {
-        $scope.forgotSuccessful = false;
+    function removeSucceed(forgotEmail) {
+      if (vm.forgotSuccessful) {
+        vm.forgotSuccessful = false;
```
andresmoschini: _If_ is not necessary here.

andresmoschini: And why does it receives an email? and it is really necessary to call _removeSucceed_ from the view? could not the validation do it automatically?

src/wwwroot/controllers/LoginCtrl.js
```diff
@@ -5,65 +5,95 @@
...
+    function setWidgetId (widgetId) {
+        vm.widgetId = widgetId;
     };
+    
+    function reloadCaptcha () {
+        vcRecaptchaService.reload(vm.widgetId);
+        vm.response = null;
+    };
+
+    function setResponse (response) {
```
andresmoschini: Maybe _setCaptchaResponse_ could be helpful?

I would like to make more explicit the flow, right now it is difficult to understand how events are being executed.

I think that it is: `submitForgot -> setResponse -> forgotPassword -> reloadCaptcha` but it is not so clear. I am not sure if it is possible to represent it in a better way.

PR #182, opened 2017-12-15
Dr-1498 Add login admin page

mvillaseco:
# Overview
 Add Login Admin Page 

# Changes
 - Add new page
 - Add new content

# In progress
 - Final endpoint
 - Tests

### DIFF
--- a/src/spec/login_spec.js
+++ b/src/spec/login_spec.js
@@ -35,4 +35,46 @@ describe('Login page', () => {
     // Assert
     expect(loginPage.getSwitchLanguageMessage()).toBe(inSpanish);
   });
+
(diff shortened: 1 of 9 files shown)


andresmoschini:
👍  We could start sharing it with @Juampi1790  to get some feedback

mvillaseco:
User not allowed to impersonate:


Client not exist:

andresmoschini:
👍

alegarciamina:
Looks great!

Just one comment, what do you think about adding another error to let the user know when client is inactive?
We could check `pending_activation` flag so we avoid the unexpected error.

PR #183, opened 2017-12-18
[DR-1472] Add new global function to load limits when user logs in

mvillaseco:
# Overview
 Load limits when the user logs in or get into reports page.

# Changes
 - Add global function


andresmoschini:
❤️

PR #192, opened 2018-01-03
Dr-1506 Dkim warnings and content change

mvillaseco:
# Overview 
 Add a new header, content change and welcome popup change.

# Changes 
 - Welcome flow.
 - New header.
 - New Label inside dkim config page.


src/wwwroot/services/auth.js
```diff
@@ -251,7 +251,8 @@
           daily: mapLimit(response.data.daily),
           hourly: mapLimit(response.data.hourly),
           hasLimits: !response.data.noLimits,
-          endDate: mapDate(response.data.endDate)
+          endDate: mapDate(response.data.endDate),
+          hasDkimReady: response.data.dkimConfigurationReady
```
andresmoschini: The name of the field is `dkimConfigurationRequired`.

You can fix it inverting the logic, something like:

```
    hasDkimReady: !response.data.dkimConfigurationRequired
```

But I think that it is better to keep the same meaning, for example:


```
    requiresDkimConfiguration: !!response.data.dkimConfigurationRequired
```

andresmoschini:
Nice work @mvillaseco !

Take into account that backend returned field is different, see https://github.com/MakingSense/Doppler_Relay/pull/1191/files#diff-8c5ac9c60c37f5a88c8e00e6706f8f0cR135

alegarciamina:
👍

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -42,6 +42,7 @@
     
     $rootScope.freeTrialStatus = null;
     $rootScope.accountLimits = { };
+    $rootScope.requiresDkimConfiguration = false;
```
andresmoschini: Why you do not directly use $rootScope.accountLimits.requiresDkimConfiguration?

andresmoschini:
👍

PR #193, opened 2018-01-03
[DR-1515] Call loadLimits after load and after login

andresmoschini:
It tries to fix the issue with showing obsolete limits data, could you review?

I did not test it yet, but I will directly test it in QA environment.


mvillaseco:
it seems ok 👍

PR #196, opened 2018-01-12
Dr-1507 domain manager improvements

mvillaseco:
# Overview
 Changes:
 - Anchor in domains
 - Content changes
 - Apply breadcrumb
 - Change submenues

# In progress
 - Tests


src/wwwroot/directives/breadcrumb.js.orig
```diff
@@ -0,0 +1,47 @@
+(function () {
```
alegarciamina: Seems you unintentionally include this merge file

mvillaseco: good catch!

alegarciamina:
Nice work!

PR #198, opened 2018-01-18
[DR-1525] Fix issues with menues and submenues

mvillaseco:
# Overview
 Fix issues with menu and Submenu

# Changes
 - Change submenu position.
 - Add underline for menues.
 - Add style

### DIFF
--- a/src/wwwroot/controllers/MainCtrl.js
+++ b/src/wwwroot/controllers/MainCtrl.js
@@ -195,6 +195,10 @@
     $scope.isPermanentAuthed = function () {
       return auth.isAuthed() && !auth.isTemporarilyAuthed();
     };
+
+    $scope.applyActive = function (name) {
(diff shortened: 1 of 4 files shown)


src/wwwroot/controllers/MainCtrl.js
```diff
@@ -195,6 +195,10 @@
     $scope.isPermanentAuthed = function () {
       return auth.isAuthed() && !auth.isTemporarilyAuthed();
     };
+
+    $scope.applyActive = function (name) {
+      return $location.path().split(/[\s/]+/).pop() == name ? true : false;
```
andresmoschini: In general, I prefer to avoid ternary operator for boolean results. Do not change it now, but what do you think about this format for future scenarios?

```csharp
$scope.applyActive = function (name) {
    return $location.path().split(/[\s/]+/).pop() == name;
```

PR #200, opened 2018-01-19
Releases/v1.0.0 beta43

alegarciamina:
- [DR-1503](http://jira.makingsense.com/browse/DR-1503) (#197) Google Analytics not working correctly
- [DR-1507](http://jira.makingsense.com/browse/DR-1507) (#195, #196) Webapp: Domain manager page improvements
- [DR-1525](http://jira.makingsense.com/browse/DR-1525) (#198, #199) Webapp: Header active state not working


src/wwwroot/app.js
```diff
@@ -123,8 +123,8 @@
           controller: 'DomainManagerCtrl',
           controllerAs: 'vm'
         })
-        .when('/settings/dkim-configuration-help', {
-          templateUrl: 'partials/settings/dkim-configuration-help.html',
+        .when('/settings/domain-manager/dkim-configuration-help', {
+          templateUrl: 'partials/settings/domain-manager/dkim-configuration-help.html',
```
andresmoschini: What does happen with the old URL? I mean, suppose that someone has created a bookmark or if we are exposing the old URL in some documentation post. Please, ensure to redirect old URL to the new one.

andresmoschini: 💤 💤 💤

andresmoschini: @mvillaseco 💤 💤 💤

andresmoschini: Fixed (I think) in #202

andresmoschini: It is supposed that this page should be accessed directly, so I will revert the "fix".

If someone has a bookmark, it will be broken, but I think that no one has a bookmark.

PR #202, opened 2018-01-23
Keep backward compatibility with deprecated URL

andresmoschini:
@mvillaseco could you review it?


andresmoschini:
It is supposed that this page should be accessed directly, so I will revert the "fix".

If someone has a bookmark, it will be broken, but I think that no one has a bookmark.

PR #206, opened 2018-02-16
[DR-1537] Handle Tracking and Dkim by using dkim flag

mvillaseco:
# Overview 
 Use Dkim flag to show if the user needs to configure it together with tracking.

# Changes:
 - Change content and remove not used.
 - Remove Tracking individual header.


andresmoschini:
Great!, feel free to replace `dkimConfigurationRequired` by `domainConfigurationRequired` or keep this change for a further release.

PR #208, opened 2018-02-22
[DR-1545] Add directive to remove not available links and redirect routing to allowed pages

mvillaseco:
# Overview
 Task: We need to add multi account users into the application that can only view certain pages.
 PR: Remove links that are not allowed to users with profile permissions and redirect them in case of a not allowed page access.

# Changes:
 - Add directive to handle links.
 - Add redirect in case of access into not allowed page.

# In Progress
 - Tests

### DIFF
--- a/src/wwwroot/app.js
+++ b/src/wwwroot/app.js
@@ -189,7 +189,6 @@
     'jwtHelper', 
     '$locale',
     'utils',
-    '$route',
     function (
       $rootScope,
       auth,
@@ -198,8 +197,7 @@
       $translate, 
       jwtHelper,
       $locale,
-      utils,
-      $route) {
+      utils) {
 
     function applyCultureFormats() {
       var locale = getLocale($translate.use());
@@ -240,7 +238,7 @@
         }
       }
 
-      verifyAuthorization($location, auth, $route);
+      verifyAuthorization($location, auth);
     });
   }]);
 
@@ -254,7 +252,7 @@
   //  });
   //}]);
 
-  function verifyAuthorization($location, auth, $route) {
+  function verifyAuthorization($location, auth) {
     var openForAllUrls = ['/signup/error', '/temporal-token-error', '/dkim-configuration-tutorial'];
     var requireLogoutUrls = ['/signup/confirmation', '/login', '/signup/registration', '/signup/succeed', '/loginAdmin'];
     var requireTemporalAuthUrls = ['/reset-password', '/change-email'];
@@ -268,7 +266,7 @@
     var pageRequireTemporalAuth = requireTemporalAuthUrls.includes(currentPath);
         
     if(!auth.isUrlAllowed(currentPath)) {
-      $location.path(auth.getDefaultUrl() || $route.routes[null].originalPath);
+      $location.path(auth.getDefaultUrl() || '/login');
     }
 
     if (pageOpenForAll) {
--- a/src/wwwroot/app.js
+++ b/src/wwwroot/app.js
@@ -268,7 +268,7 @@
     var pageRequireTemporalAuth = requireTemporalAuthUrls.includes(currentPath);
         
     if(!auth.isUrlAllowed(currentPath)) {
-      $location.path($route.routes[auth.getDefaultUrl()].originalPath);
+      $location.path(auth.getDefaultUrl() || $route.routes[null].originalPath);
     }
 
     if (pageOpenForAll) {
--- a/src/wwwroot/app.js
+++ b/src/wwwroot/app.js
@@ -266,10 +266,9 @@
     var pageOpenForAll = openForAllUrls.includes(currentPath);
     var pageRequireLogout = requireLogoutUrls.includes(currentPath);
     var pageRequireTemporalAuth = requireTemporalAuthUrls.includes(currentPath);
-    var tokenPermissons = auth.decodeToken();
         
-    if(tokenPermissons && !currentPath.match(tokenPermissons.acceptedUrlsPattern)){
-      $location.path($route.routes[tokenPermissons.defaultUrl].originalPath);
+    if(!auth.isUrlAllowed(currentPath)) {
+      $location.path($route.routes[auth.getDefaultUrl()].originalPath);
     }
 
     if (pageOpenForAll) {
--- a/src/wwwroot/app.js
+++ b/src/wwwroot/app.js
@@ -189,6 +189,7 @@
     'jwtHelper', 
     '$locale',
     'utils',
+    '$route',
     function (
       $rootScope,
       auth,
@@ -197,7 +198,8 @@
       $translate, 
       jwtHelper,
       $locale,
-      utils) {
+      utils,
+      $route) {
 
     function applyCultureFormats() {
       var locale = getLocale($translate.use());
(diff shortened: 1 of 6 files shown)


mvillaseco:
# Cases:

 # Reports:


 # Templates:


 # Settings:

src/wwwroot/services/auth.js
```diff
@@ -33,30 +33,58 @@
...
       temporarilyAuthed = !!isTemporal;
       if (!temporarilyAuthed) {
         $window.localStorage.setItem('jwtToken', token);
       }
     }
 
+    function decodeToken() {
+      decodedToken = jwtHelper.decodeToken(encodedToken);
+      if (!decodedToken.profile) {
+        return permissons = null;
```
andresmoschini: I think that it is not a good idea to hay a function that returns value and also generates side effects. 

What do you think about considering _permissons_ as part of decoded token?

```js
function decodeToken() {
  decodedToken = jwtHelper.decodeToken(encodedToken);
  decodedToken.permissons = expandPermissons(decodedToken.profile);
}

function expandPermissons(profile) {
  // If switch works well with null, this condition could be removed
  if (!profile) {
    return null;
  }
  switch (profile) {
    case "reports": return {
      acceptedUrlsPattern: /^#?\/reports\/?(\?.*)?$|^#?\/reports\/downloads\/?(\?.*)?$/,
      defaultUrl: "/reports"
    };
    case "templates": return {
      acceptedUrlsPattern: /^#?\/templates\/?(\?.*)?$/,
      defaultUrl: "/templates"
    };
    case "settings": return {
      acceptedUrlsPattern: /^#?\/settings\//,
      defaultUrl: "/settings/connection-settings"
    };
    default: return null;
}
```

src/wwwroot/app.js
```diff
@@ -264,6 +266,11 @@
     var pageOpenForAll = openForAllUrls.includes(currentPath);
     var pageRequireLogout = requireLogoutUrls.includes(currentPath);
     var pageRequireTemporalAuth = requireTemporalAuthUrls.includes(currentPath);
+    var tokenPermissons = auth.decodeToken();
```
andresmoschini: I think that _tokenPermissons_ could be considered part of _decodedToken_ I would not add more calls to _decodeToken()_ than actually exists.

src/wwwroot/directives/link-toggler.js
```diff
@@ -0,0 +1,34 @@
...
+  
+      linkToggler.$inject = [
+      "auth"
+    ];
+  
+    function linkToggler(auth) {
+      var directive = {
+        restrict: 'A',
+        link: function (scope, element, attrs, ctrls) {
+            var decodedTkn = auth.decodeToken();
```
andresmoschini: I think that you should avoid to call _decodeToken()_.

Permissons object should be already available.

src/wwwroot/services/auth.js
```diff
@@ -136,12 +164,14 @@
...
         encodedToken = token;
-        decodedToken = jwtHelper.decodeToken(encodedToken);
+        decodeToken();
       }
     }
 
+
     function logOut() {
       decodedToken = null;
+      permissons = null;
```
andresmoschini: I think think that permissons object could be considered as part of _decodedToken_ but it is a valid approach to have an independent variable for it.

andresmoschini:
You are the best @mvillaseco !

Your _link-toggler_ directive seems to be very clever and useful. 

I am CCing some people that could be interested on it or give us some advises: @cbernat, @stirelli, @juanasas, @morenoisidro, @lassmann

src/wwwroot/directives/link-toggler.js
```diff
@@ -13,19 +13,16 @@
...
-                hrefAllowed = attrs.href.match(decodedTkn.acceptedUrlsPattern) ? true : false;
-              }
-              if(!hrefAllowed) {
-                element.remove();
-              }
-            }
+          var hrefAllowed;
+          if(!attrs.href) {
+            // Finds the first element inside the container and analize.
+            hrefAllowed = auth.isUrlAllowed(element[0].querySelector('a').href.split("#")[1]);
```
andresmoschini: Did you test it with multiples `a` elements?

src/wwwroot/services/auth.js
```diff
@@ -34,11 +34,12 @@
...
       addFreeTrialNotificationToStorage: addFreeTrialNotificationToStorage,
       changeEmail: changeEmail,
-      decodeToken: decodeToken
+      isUrlAllowed: isUrlAllowed,
+      getDefaultUrl: getDefaultUrl
     };
 
     var decodedToken = null;
-    var permissons = null;
+    var permissions = null;
```
andresmoschini: I think that you are not using this variable anymore

src/wwwroot/services/auth.js
```diff
@@ -171,7 +179,7 @@
 
     function logOut() {
       decodedToken = null;
-      permissons = null;
+      permissions = null;
```
andresmoschini: You are not using this variable anymore.

src/wwwroot/app.js
```diff
@@ -266,10 +266,9 @@
     var pageOpenForAll = openForAllUrls.includes(currentPath);
     var pageRequireLogout = requireLogoutUrls.includes(currentPath);
     var pageRequireTemporalAuth = requireTemporalAuthUrls.includes(currentPath);
-    var tokenPermissons = auth.decodeToken();
         
-    if(tokenPermissons && !currentPath.match(tokenPermissons.acceptedUrlsPattern)){
-      $location.path($route.routes[tokenPermissons.defaultUrl].originalPath);
+    if(!auth.isUrlAllowed(currentPath)) {
+      $location.path($route.routes[auth.getDefaultUrl()].originalPath);
```
andresmoschini: I know that it does not happen, but since it is legal that _getDefaultUrl()_ returns _null_ I think that we should take into account that case here.

andresmoschini: For example:

```js
if (!auth.isUrlAllowed(currentPath)) {
    $location.path(auth.getDefaultUrl() || '/reports');
```

Also, I am not sure why do you search the route in _$route.routes_. I mean, if the route exist it should be true_ `x
$route.routes[x].originalPath`, Right? Am I missing something?

src/wwwroot/directives/link-toggler.js
mvillaseco: yes, it will only analize the first one. (and remove or not the container if the first "A" should be removed)

andresmoschini: Talking with @mvillaseco, he explain me that _querySelector_ returns the first element that matches the query.

src/wwwroot/app.js
andresmoschini: Regarding `x == $route.routes[x].originalPath`, it is right, but when there is no default route (when `x == null`) we want to redirect to _global default URL_, @mvillaseco is changing the code to make that behavior more clear.

src/wwwroot/directives/link-toggler.js
```diff
@@ -0,0 +1,31 @@
...
+      "auth"
+    ];
+  
+    function linkToggler(auth) {
+      var directive = {
+        restrict: 'A',
+        link: function (scope, element, attrs, ctrls) {
+          var hrefAllowed;
+          if(!attrs.href) {
+            // Finds the first element inside the container and analize.
```
morenoisidro: It's really helpful to have comments like this inside the code. Thx!

PR #210, opened 2018-02-28
[DR-1546] Refresh header links before login

mvillaseco:
# Overview
 Fix issue with header links not being displayed and small css fix.


andresmoschini:
👍

PR #214, opened 2018-03-12
[DR-1551] Fix wrong regex in templates profile

mvillaseco:
# Overview:
 Fix issue with regex when user have a templates profile. Now we will allow anything after /templates/ and /templates only(without the '/')


andresmoschini:
👍 Thanks!

PR #215, opened 2018-03-13
[DR-1551] Fix issue with revision numbers replacing templates string

mvillaseco:
# Overview
It seems that we have a problem with the gulp task that adds the revision numbers, it's currently adding it into any templates string inside auth.js(maybe in all services).

This is an ugly but working fix. I add also a TODO to research this, buy for now, we can go with this solution.

### DIFF
--- a/src/wwwroot/services/auth.js
+++ b/src/wwwroot/services/auth.js
@@ -65,12 +65,14 @@
     }
 
     function expandPermissions(profile) {
+      // TODO: Review why the revision numbers are being added into the word templates in this file.
+      var TEMPLATES = " templates ".trim();
       switch (profile) {
         case "reports": return {
           acceptedUrlsPattern: /^#?\/reports\/?(\?.*)?$|^#?\/reports\/downloads\/?(\?.*)?$/,
           defaultUrl: "/reports"
         };
-        case "templates": return {
+        case TEMPLATES: return {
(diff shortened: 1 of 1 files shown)


src/wwwroot/services/auth.js
```diff
@@ -65,12 +65,14 @@
...
     function expandPermissions(profile) {
+      // TODO: Review why the revision numbers are being added into the word templates in this file.
+      var templates = " templates ";
       switch (profile) {
         case "reports": return {
           acceptedUrlsPattern: /^#?\/reports\/?(\?.*)?$|^#?\/reports\/downloads\/?(\?.*)?$/,
           defaultUrl: "/reports"
         };
-        case "templates": return {
+        case templates.trim(): return {
```
andresmoschini: What do you think about to move the patch up and also rename _templates_ as _TEMPLATES_ to make more clear yet that it is a constant and a patch?

```js
function expandPermissions(profile) {
  // TODO: Review why the revision numbers are being added into the word templates in this file.
  var TEMPLATES = " templates ".trim();
  switch (profile) {
    case "reports": return {
      acceptedUrlsPattern: /^#?\/reports\/?(\?.*)?$|^#?\/reports\/downloads\/?(\?.*)?$/,
      defaultUrl: "/reports"
    };
    case TEMPLATES: return {
      // . . .
```

PR #217, opened 2018-03-14
[Dr-1517] Add origin to registration user data

mvillaseco:
# Overview
 Add origin into registration data from user in order to use it in zoho lead. Also add origin description when the user comes from login.


src/wwwroot/partials/login/login.html
```diff
@@ -1,5 +1,5 @@
 <div class="section--wrapper second-line--header login--link">
-  <p>{{ 'dont_have_account_text' | translate }} <a href="#/signup/registration">{{ 'register_now' | translate }}</a></p>
+  <p>{{ 'dont_have_account_text' | translate }} <a href="#/signup/registration?origin=login">{{ 'register_now' | translate }}</a></p>
```
andresmoschini: Is it not possible to have origin parameter in root page?

PR #221, opened 2018-03-21
Feature/dr-1514

cristiangiagante:
# Changes
-Upgradeable validation implemented on Plan controller to DKIM and Delivered messages
-Auxiliar methods to prevent click on link buttons
-Mocked localization messages

Pending:
Tests
Frontend layout integration

### DIFF
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -25,6 +25,7 @@
       { text: 'submenu_my_profile', url: 'settings/my-profile', active: false },
       { text: 'submenu_my_plan', url: 'settings/my-plan', active: true }
     ]);
+    vm.errorPlanUpgrade = false;
     vm.hideDragMe = false;
     vm.activationPromise = activate();
     var defaultPlanDeliveries = '60000';
@@ -153,7 +154,8 @@
     }
 
     function showPricingChart() {
-        vm.pricingChartDisplayed = true;
+        vm.pricingChartDisplayed = vm.currentMonthlyCount>0? true : false;
+        vm.errorPlanUpgrade = vm.currentMonthlyCount>0? false : true;
     }
 
     function isCurrentPlan(plan){
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -61,7 +61,7 @@
       var getPlansAvailable = settings.getPlansAvailable().then(function(response) {
         planItems = response.data.items;
       });
-      return Promise.all([getPlansAvailable, getCurrentPlanInfo, getMonthConsumption()]).then(function(){        
+      return Promise.all([getPlansAvailable, getCurrentPlanInfo, getMonthConsumption(), getDeliveriesConsumption()]).then(function(){        
         ensureValidDefaultPlanDelivery();
         loadSlider();
         changePlan(defaultPlanDeliveries);        
@@ -80,6 +80,14 @@
             if (vm.currentPlanEmailsAmount < vm.currentMonthlyCount) {
               vm.extraEmailsSent = vm.currentMonthlyCount - vm.currentPlanEmailsAmount;
             }
+          });
+    }
+
+    function getDeliveriesConsumption() {
+      //Get total messages delivered by the account
+      return settings.getStatusPlanInfo()
+          .then(function (result) {
+            vm.totalDeliveriesCount = result.data.totalDeliveriesCount;
           });
     }
 
@@ -187,7 +195,7 @@
       }
       else{
         //Exist activity in the current plan?
-        errorsPlanUpgrade.isDeliveredError = vm.currentMonthlyCount > 0? false : true;
+        errorsPlanUpgrade.isDeliveredError = vm.totalDeliveriesCount > 0? false : true;
         //It is the domain correctly configurated?
         errorsPlanUpgrade.isDKIMError = limit.requiresDomainConfiguration; 
         //This account is not registering deliveries and is not correctly configurated?
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -16,16 +16,16 @@
     '$filter',
     'reports',
     'moment',
-    'utils'
+    'utils',
+    '$log'
   ];
 
-  function PlanCtrl($scope, $location, $rootScope, auth, $translate, $timeout, settings, $filter, reports, moment, utils) {
+  function PlanCtrl($scope, $location, $rootScope, auth, $translate, $timeout, settings, $filter, reports, moment, utils, $log) {
     var vm = this;
     $rootScope.setSubmenues([
       { text: 'submenu_my_profile', url: 'settings/my-profile', active: false },
       { text: 'submenu_my_plan', url: 'settings/my-plan', active: true }
     ]);
-    vm.errorPlanUpgrade = false;
     vm.hideDragMe = false;
     vm.activationPromise = activate();
     var defaultPlanDeliveries = '60000';
@@ -36,7 +36,8 @@
     vm.planInfoLoader = true;
     vm.planStatusInfoLoader = true;
     vm.ipsPlanCount = 0;
-
+    vm.errorsPlanUpgrade = isUpgradeablePlan(); //DR-1514
+    vm.isValidUpgradeablePlan = isValidUpgradeablePlan(); //DR-1514
     function activate() {
       var getCurrentPlanInfo = settings.getCurrentPlanInfo().then(function(response) {
         vm.currentPlanPrice = response.data.fee + (response.data.ips_count * response.data.cost_by_ip || 0);
@@ -154,8 +155,7 @@
     }
 
     function showPricingChart() {
-        vm.pricingChartDisplayed = vm.currentMonthlyCount>0? true : false;
-        vm.errorPlanUpgrade = vm.currentMonthlyCount>0? false : true;
+        vm.pricingChartDisplayed = true;
     }
 
     function isCurrentPlan(plan){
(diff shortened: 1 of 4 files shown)


src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -81,6 +83,14 @@
           });
     }
 
+    function getDeliveriesConsumption() {
```
mvillaseco: I think this function it's not needed, the exact same service call is made in getCurrentPlanInfo, maybe rename that function or add a boolean to check if the deliveries are more than X amount (consider that this call it's going to be really slow).

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -35,7 +36,8 @@
     vm.planInfoLoader = true;
     vm.planStatusInfoLoader = true;
     vm.ipsPlanCount = 0;
-
+    vm.errorsPlanUpgrade = isUpgradeablePlan(); //DR-1514
```
mvillaseco: Im not sure why we need to have the Jira ticket number in the code.

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +184,42 @@
       }      
     }
 
+    function isUpgradeablePlan(limit){
+      //Check constraints to validate that plan is upgradeable
+
+      //Set to true all errors to prevent access if service is not working
+      var errorsPlanUpgrade = {'isDKIMError': true, "isDeliveredError": true, "isDeliveredAndDKIMError": true};
```
mvillaseco: There is already a global variable to check if there is a Dkim, or domain error to be configured, it's currently being used in the header, review the function called loadLimits in MainCtrl.

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +184,42 @@
...
+
+      if(limit==undefined){
+        var isConfigRequired = auth.getLimitsByAccount().then(isUpgradeablePlan).catch($log.error);
```
mvillaseco: i dont think $log is useful to anyone, if any service fails there is already a ErrorHandler file to catch and show an error in every specific case.

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +184,42 @@
...
+
+      return errorsPlanUpgrade;
+    }
+
+    function isValidUpgradeablePlan(){
+      var haveErrors = (vm.errorsPlanUpgrade.isDKIMError || vm.errorsPlanUpgrade.isDeliveredError || vm.errorsPlanUpgrade.isDeliveredAndDKIMError)? true : false;
+      return !haveErrors;
+    }
+
+    $scope.disableNotUpgradable=function(event){
```
mvillaseco: Check that there are controllers with Sintax "Controller as vm", there are several benefits from using "scope".

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +184,42 @@
...
+      if(vm.isValidUpgradeablePlan){
+        return true;
+      }
+      else{
+        event.preventDefault();
```
mvillaseco: I think it's better to not allow the user to do an action before, rather stopping an action after doing it.

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +184,42 @@
       }      
     }
 
+    function isUpgradeablePlan(limit){
```
mvillaseco: We are trying to use https://google.github.io/styleguide/javascriptguide.xml -> "Method and property definitions" spacing

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +184,42 @@
...
+
+      //Set to true all errors to prevent access if service is not working
+      var errorsPlanUpgrade = {'isDKIMError': true, "isDeliveredError": true, "isDeliveredAndDKIMError": true};
+
+      if(limit==undefined){
+        var isConfigRequired = auth.getLimitsByAccount().then(isUpgradeablePlan).catch($log.error);
+      }
+      else{
+        //Exist activity in the current plan?
+        errorsPlanUpgrade.isDeliveredError = vm.totalDeliveriesCount > 0? false : true;
```
andresmoschini: Very minor, but

```js
errorsPlanUpgrade.isDeliveredError = vm.totalDeliveriesCount > 0? false : true;
``` 

Is the same than:

```js
errorsPlanUpgrade.isDeliveredError = vm.totalDeliveriesCount <= 0;
```

or
 
```
errorsPlanUpgrade.isDeliveredError = !(vm.totalDeliveriesCount > 0);
```

src/wwwroot/controllers/PlanCtrl.js
andresmoschini: if we want to use _DOM_, we could disable the button. If we want to use logic, we do not need to patch the event handling.

andresmoschini:
Great work @cristiangiagante! I think that it is not exactly what we want but it is a nice approximation to start with.

I understand that you will rewrite it without paying attention to the deliveries count by the moment and create a new PR. Thanks!

src/wwwroot/partials/settings/my-plan.html
```diff
@@ -106,8 +106,8 @@ <h5 class="special uppercase">{{ 'my_plan_pricing_chart' | translate }}</h5>
         </li>
         <li class="strong">
           <h2 id="planPriceBig" class="special plan--price-big">{{vm.currency | translate}} {{vm.leftPlanPrice | number:2 | numberFormat}}</h2>
-          <a ng-class="!vm.disableLeftPlan ? 'show':''" class="hidden-feature-by-display button button--small" href="#/settings/billing?plan={{vm.leftPlanName}}">{{'change_plan' | translate}}</a>
-          <a ng-class="vm.disableLeftPlan ? 'show':''" class="button button--small button--disabled hidden-feature-by-display">{{'change_plan' | translate}}</a>
+          <a ng-class="[!vm.disableLeftPlan ? 'show':'', vm.isValidUpgradeablePlan ? '':'button--disabled']" ng-click="disableNotUpgradable($event)" class="hidden-feature-by-display button button--small" href="#/settings/billing?plan={{vm.leftPlanName}}">{{'change_plan' | translate}}</a>
```
morenoisidro: Hey! I'm pretty new here... So, sorry if there's already been a discussion on this topic; just bringing the idea: Would it be worth having values like `"#/settings/billing?plan={{vm.leftPlanName}}"` being handled from the handled from the JS side? That would give you  the chance to externalize common route values to variables. Then you can put the string value in a variable and just add it to `href` attribute as an angular expression.

morenoisidro: I usually tend to avoid using logic in the HTML (when it can be done... of course), and try to put things in a more declarative way. But that is quite personal style... It also make debugging a little easier. Just sharing thoughts.

andresmoschini: Thanks @morenoisidro ! I am really glad that you are reviewing our repo.

I have not an strong opinion on it, @mvillaseco what do you think?

morenoisidro: Moving conversation to a separate thread.

I usually tend to avoid using logic in the HTML (when it can be done... of course), and try to put things in a more declarative way.
Like a function call, instead of [!vm.disableLeftPlan ? 'show':'', vm.isValidUpgradeablePlan ? '':'button--disabled']
But that is quite personal style... It also make debugging a little easier. Just sharing thoughts.

morenoisidro: WC!

mvillaseco: Thanks for sharing your thoughts!, will be considered.

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Deliveries implementation will be removed until decide the best approach to retrieve data from db. I will send the next commit without this. Thanks @mvillaseco

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Will be deleted. I am currently using it to quickly recognize my implementations. Thanks @mvillaseco

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Thanks @mvillaseco I will check it and fix up this 👍

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Thanks @mvillaseco I will apply this change in the next commit.

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: @andresmoschini this was talked with @mvillaseco, currently the buttons are not really buttons, they are <a> tags being formatted to a button with css.

I apply this approach to prevent to change the html structure. To not used this I need to create new styles to prevent add bugs in styles to other elements that could be using the same approach using <a> elements.

Mauro recommend me to use pointer-events: none to avoid the js to prevent the action. But this could have problems with cross-platform compatibility following this table:
git

Let me know what do you think about mantain this approach or change it for another one.

Thanks!

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Perfect, I will read it.

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Thanks I will implement it.

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: UPDATE: I will implement current logic to disabled buttons instead to add a new inline validation.

andresmoschini: Thanks for your research @cristiangiagante, at the end we have talked and decided to try to reproduce the old approach, I mean, the same that we are doing to disable current plan to not allow to select it again.

image

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: I will change this approach to custom events to get existing limit status information when is already loaded from PlanController

PR #222, opened 2018-03-26
Feature/dr 1561 require ready domains

cristiangiagante:
## Background
  
Integrate Frontend changes on rail-webapp to show validation requirements in plans modal
  
## Changes done
  
Styles to new content box added
Localization languages added
Improvements on front end logic related to code review (https://github.com/DopplerRelay/relay-webapp/pull/221)
  
## Pending to be done
  
Deliveries validation
  
## Notes
Related Mockups:

https://projects.invisionapp.com/d/main#/console/11110634/286241939/preview

https://projects.invisionapp.com/d/main#/console/11110634/286241938/preview
## Demo
 
Tested in Edge and Chrome

### DIFF
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -35,6 +35,8 @@
     vm.planInfoLoader = true;
     vm.planStatusInfoLoader = true;
     vm.ipsPlanCount = 0;
+    vm.upgradePlanErrors = {'isDomainReady' : false};
+    vm.isValidUpgradeablePlan = isValidUpgradeablePlan;
 
     function activate() {
       var getCurrentPlanInfo = settings.getCurrentPlanInfo().then(function(response) {
@@ -174,6 +176,12 @@
       }      
     }
 
+    function isValidUpgradeablePlan() {
+      if($rootScope.accountLimits.requiresDomainConfiguration!=undefined)
+        vm.upgradePlanErrors.isDomainReady = !$rootScope.accountLimits.requiresDomainConfiguration;
+      var isValid=vm.upgradePlanErrors.isDomainReady;
+      return isValid;
+    }
   }
 
 })();
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -35,9 +35,8 @@
     vm.planInfoLoader = true;
     vm.planStatusInfoLoader = true;
     vm.ipsPlanCount = 0;
-    vm.upgradePlanErrors = {'isDomainReady' : false};
     vm.isValidUpgradeablePlan = isValidUpgradeablePlan;
-
+    vm.requiresDomainConfiguration = requiresDomainConfiguration;
     function activate() {
       var getCurrentPlanInfo = settings.getCurrentPlanInfo().then(function(response) {
         vm.currentPlanPrice = response.data.fee + (response.data.ips_count * response.data.cost_by_ip || 0);
@@ -177,10 +176,11 @@
     }
 
     function isValidUpgradeablePlan() {
-      if($rootScope.accountLimits.requiresDomainConfiguration!=undefined)
-        vm.upgradePlanErrors.isDomainReady = !$rootScope.accountLimits.requiresDomainConfiguration;
-      var isValid=vm.upgradePlanErrors.isDomainReady;
-      return isValid;
+      return !requiresDomainConfiguration();
+    }
+  
+    function requiresDomainConfiguration() {
+      return $rootScope.accountLimits.requiresDomainConfiguration || false;
     }
   }
 
--- a/src/wwwroot/styles/modules/_button.scss
+++ b/src/wwwroot/styles/modules/_button.scss
@@ -11,6 +11,12 @@ $input--color: #525845;
 $input--error-background: #DF3434;
 $input--error--color: #FFF;
 
+$arrowbox--error-color: #e78224;
+$arrowbox--error-background: #ffffff;
+$arrowbox--error-font: #000000;
+$arrowbox--error-title: #777777;
+$arrowbox--list-color: #e78224;
+$arrowbox--list-link: #e78224;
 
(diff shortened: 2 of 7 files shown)


andresmoschini:
@cristiangiagante could you include a couple of screenshots?

andresmoschini:
@cristiangiagante, just to confirm, all these commits have sense in this PR (they seems to be a lot)?

image

andresmoschini:
> @cristiangiagante, just to confirm, have all these commits sense in this PR (they seems to be a lot)?

It seems that no because there are changes related to _origin_ parameter. Please, rebase it to have a clean PR with only your changes.

src/wwwroot/controllers/MainCtrl.js
```diff
@@ -189,6 +190,9 @@
     $rootScope.$on('$locationChangeStart', function () {
       submenues = [];
     });
+    $rootScope.getLoadedLimits = function () {
+        $rootScope.$broadcast('LimitsLoaded');
```
andresmoschini: I think that we do not need it, but, if we really need it could be done in the same place where _$rootScope.accountLimits_ is updated without using _watch_

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -35,7 +35,8 @@
     vm.planInfoLoader = true;
     vm.planStatusInfoLoader = true;
     vm.ipsPlanCount = 0;
-
+    vm.errorsPlanUpgrade = UpgradeableStatusPlan();
```
andresmoschini: I do not understand the meaning of this line.

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -35,7 +35,8 @@
...
+    vm.isValidUpgradeablePlan = isValidUpgradeablePlan();
```
andresmoschini: It is evaluating once when this code is executed, so you are loosing the databinding, is it fine?

src/wwwroot/styles/modules/_button.scss
```diff
@@ -11,6 +11,12 @@ $input--color: #525845;
 $input--error-background: #DF3434;
 $input--error--color: #FFF;
 
+$arrowbox--error-color:           #e78224;
+$arrowbox--error-background:      #ffffff;
+$arrowbox--error-font: 			  #000000;
+$arrowbox--error-title:			  #777777;
+$arrowbox--list-color: 			  #e78224;
+$arrowbox--list-link:   		  #e78224;
```
andresmoschini: Something happened with the indentation

andresmoschini:
Hi @cristiangiagante, nice changes, at the end it was not so big, and it is easy to review. Maybe we could make a call to review the issues updating limits.

Could you also include some screenshots in the PR description?

cristiangiagante:
@andresmoschini  I refreshed the history of my branch to be more cleanest.
I attached some images:

**Domain required not Ready**
image

**Domain required not Ready contact button**
image

**Domain required Ready**
image

andresmoschini:
Thanks @cristiangiagante, the screenshots look great. Since we will finish our sprint tomorrow, and probably we will no have time to finish the second validation (having at least one delivery) properly, do you think that current content appropriate as it is? or should we make some cahnges?

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -174,6 +176,12 @@
       }      
     }
 
+    function isValidUpgradeablePlan() {
+      if($rootScope.accountLimits.requiresDomainConfiguration!=undefined)
+        vm.upgradePlanErrors.isDomainReady = !$rootScope.accountLimits.requiresDomainConfiguration;
+      var isValid=vm.upgradePlanErrors.isDomainReady;
+      return isValid;
```
andresmoschini: Please, revert this refactoring.

Based on the name, _isValidUpgradeablePlan()_ seems to be a function that is useful to know a particular state: _if the account is in a valid state to allow upgrade the agreement_, right?

But, it also has side effects that are not clear in the function name. In general is not a good idea having functions with side effects.

In this particular case, the problem is that the value of _vm.upgradePlanErrors_ will be only updated if we have executed _isValidUpgradeablePlan()_ before.

Does it make sense?

cristiangiagante:
Thanks for you comments @andresmoschini I think it is ok, we can move to QA and do a better analysis of how to implement deliveries, I have some ideas to share.

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: Yes I understand @andresmoschini, I add this behavior to not depend only from backend requests. With this object we have a default value to validate, In the other hand with the other approach we get undefined values until the $rootScope.accountLimits.requiresDomainConfiguration is populated with data from the api.

andresmoschini: Ok, but a function that seems to simply return a value that has side effects is not acceptable, could you think in another alternative?

andresmoschini: > With this object we have a default value to validate, In the other hand with the other approach we get undefined values until the $rootScope.accountLimits.requiresDomainConfiguration is populated with data from the api.

That is not a problem, when it is undefined you can decide the default value

cristiangiagante: Perfect @andresmoschini I will change this approach

src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -35,7 +35,9 @@
     vm.planInfoLoader = true;
     vm.planStatusInfoLoader = true;
     vm.ipsPlanCount = 0;
-
+    vm.upgradePlanErrors = {'isDomainReady' : false};
```
andresmoschini: I think that this line is not necessary anymore.

andresmoschini:
Nice work @cristiangiagante! thanks for your patience, please remove the unnecessary line and :shipit: !

PR #223, opened 2018-03-28
Feature/dr-1561-require ready domains

cristiangiagante:
## Background
  
Results from test on QA
  
## Changes done
  
Some logic in the view was fixed to prevent show two buttons in the same side of plan subscription
  
## Pending to be done
  
N/A
  
## Notes
  
Add a new test to check that only must exists 1 button in each side of plan subscription containers
 
## Demo
 ISSUE:
image

FIX:
image


andresmoschini:
👍

PR #225, opened 2018-03-28
Feature/dr 1561 require ready domains

cristiangiagante:
## Background
  
Styles issue detected testing on QA
  
## Changes done
-Add require domain validation in plans view
-Styles moved to _settings.scss
-Fix styles to be responsive with mobile resolutions
  
## Notes
Tested on Chrome, Edge, Firefox (last versions) in desktop and mobile mode.  

## Demo
 ISSUE:
image

FIX:
image


src/wwwroot/styles/views/_settings.scss
```diff
@@ -620,6 +620,67 @@
...
+      bottom: 100%;
+      left: 24%;
+      border: solid transparent;
+      content: " ";
+      position: absolute;
+      border-bottom-color: $suit-accent-color;
+      border-width: 9px;
+      margin-left: -9px;
+    }
+    ol{
```
mvillaseco: ol is for ordered lists, im not sure if this is the case, maybe UL is more common to use in this cases, just consider it

src/wwwroot/styles/views/_settings.scss
```diff
@@ -620,6 +620,67 @@
...
+      padding-left: 15px;
+      margin-bottom: 0px;
+      li{
+        border-left: 0px;
```
mvillaseco: those borders :(, if it has to be like this, then leave it, we can fix that later.

src/wwwroot/styles/views/_settings.scss
cristiangiagante: Is an orderer list based on the mocks.

src/wwwroot/styles/views/_settings.scss
cristiangiagante: I attached an image to explain why I need to override this value explicitly.
image

PR #228, opened 2018-04-05
[DR-1578] Create deliveries validation in frontend

cristiangiagante:
## Background
  
Update logic and plan view to validate using deliveries constraint
  
## Changes done
  
* Add new property from api
* Add localization messages
* Update styles
* Add tests

## Pending to be done
N/A

## Notes

I update the styles due to pseudo element last-of-type was being applied only in the last <li> element, that was ok for 1 message but now we have a max of two messages 2, with the new styles the box support unlimited number of messages and without bugs.

### DIFF
--- a/src/wwwroot/controllers/PlanCtrl.js
+++ b/src/wwwroot/controllers/PlanCtrl.js
@@ -37,6 +37,7 @@
     vm.ipsPlanCount = 0;
     vm.isValidUpgradeablePlan = isValidUpgradeablePlan;
     vm.requiresDomainConfiguration = requiresDomainConfiguration;
+    vm.requiresDeliveries = requiresDeliveries;
     function activate() {
       var getCurrentPlanInfo = settings.getCurrentPlanInfo().then(function(response) {
         vm.currentPlanPrice = response.data.fee + (response.data.ips_count * response.data.cost_by_ip || 0);
@@ -176,11 +177,15 @@
     }
 
     function isValidUpgradeablePlan() {
-      return !requiresDomainConfiguration();
+      return !requiresDomainConfiguration() && !requiresDeliveries();
     }
   
     function requiresDomainConfiguration() {
-      return $rootScope.accountLimits.requiresDomainConfiguration || false;
+      return !!$rootScope.accountLimits.requiresDomainConfiguration;
+    }
+
+    function requiresDeliveries(){
+      return !!$rootScope.accountLimits.requiresDeliveries;
     }
   }
 
--- a/src/wwwroot/services/auth.js
+++ b/src/wwwroot/services/auth.js
@@ -291,7 +291,8 @@
           hourly: mapLimit(response.data.hourly),
           hasLimits: !response.data.noLimits,
           endDate: mapDate(response.data.endDate),
-          requiresDomainConfiguration: !!response.data.domainConfigurationRequired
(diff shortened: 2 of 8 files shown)


src/wwwroot/controllers/PlanCtrl.js
```diff
@@ -176,12 +177,16 @@
...
-      return !requiresDomainConfiguration();
+      return !requiresDomainConfiguration() && !requiresDeliveries();
     }
   
     function requiresDomainConfiguration() {
       return $rootScope.accountLimits.requiresDomainConfiguration || false;
     }
+
+    function requiresDeliveries(){
+      return $rootScope.accountLimits.requiresDeliveries || false;
```
mvillaseco: the same as !!$rootScope.accountLimits.requiresDeliveries? (not sure)

src/wwwroot/services/auth.js
```diff
@@ -291,7 +291,8 @@
           hourly: mapLimit(response.data.hourly),
           hasLimits: !response.data.noLimits,
           endDate: mapDate(response.data.endDate),
-          requiresDomainConfiguration: !!response.data.domainConfigurationRequired
+          requiresDomainConfiguration: !!response.data.domainConfigurationRequired,
+          requiresDeliveries: response.data.hasNotDeliveries
```
mvillaseco: Maybe you could handle the || false here instead of the controller.

src/wwwroot/styles/views/_settings.scss
```diff
@@ -654,14 +654,13 @@
       li{
         border-left: 0px;
         border-right: 0px;
+        border-bottom: 0px;
+        padding: 0px;
```
mvillaseco: This is not looking good 🤕

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: I will check it thanks for the review :)

src/wwwroot/services/auth.js
cristiangiagante: I will do it :) thanks

src/wwwroot/styles/views/_settings.scss
cristiangiagante: I will removed them in the next update 👍 thanks for your help Mauro

src/wwwroot/controllers/PlanCtrl.js
cristiangiagante: But we need handle this in the planController too because is called first than the auth service (async request take a seconds to response with popupated data). We catch the undefined and convert it to false when the PlanController is initialized. I will use !!$rootScope.accountLimits.requiresDeliveries instead the current sintax. Thanks!

src/wwwroot/styles/views/_settings.scss
```diff
@@ -443,32 +443,37 @@
...
-      list-style-type: none;
-      border-left: $suit-border-default;
-      border-right: $suit-border-default;
-      padding:  0.5em 1em;
-      color:$suit-font-color-light;
-      .hidden-feature-by-display { 
-        display: none;
-        transition: none;
-        &.show{display: inline-block;}
+    ul{
```
mvillaseco: This saves so much ugly stuff :) Perfect!

PR #229, opened 2018-04-10
[DR-1589] Show limit status box if the daily limit is zero

cristiangiagante:
## Background
  
Users can't see their limit status if it's zero
  
## Changes done
  
-Update reports view to show the limit status box always
-Create tests
  
## Pending to be done
  
N/A
  
## Notes
  
N/A
 
## Demo
 
image

### DIFF
--- a/src/spec/reports_spec.js
+++ b/src/spec/reports_spec.js
@@ -240,4 +240,39 @@ describe('Reports page', () => {
     // Assert
     expect(reportsPage.isMaxRateDailyLimitDisplayed()).toEqual(false);
   });
+
+  it('should show the daily limits for accounts that has zero limits', () => {
+    // Arrange
+    beginAuthenticatedSession();
+    browser.addMockModule('descartableModule2', () => angular
+      // This code will be executed in the browser context,
+      // so it cannot access variables from outside its scope
+      .module('descartableModule2', ['ngMockE2E'])
+      .run($httpBackend => {
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/statistics\/events\/(by_hour|by_day)/).respond(200, {
+          "items": []
+        });
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/deliveries/).respond(200, {
+          "items": []
+        });
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/statistics\/deliveries\/(by_hour|by_day)/).respond(200, {
+          "items": [{ 'total': 4, 'sent': 2, 'dropped': 0 }, { 'total': 6, 'sent': 6, 'dropped': 0 }]
+        });
+        $httpBackend.whenGET(/\/accounts\/[\w|-]*\/status\/limits/).respond(200, {
+          "daily": {
+            "limit": 0,
+            "remaining": 0,
+            "reset": "2017-09-16T00:00:00Z"
+          }
+        });
+      }));
+    var reportsPage = new ReportsPage();
+
+    // Act
+    browser.get('/#/reports');
+
+    // Assert
+    expect(reportsPage.isMaxRateDailyLimitDisplayed()).toEqual(true);
+    expect(reportsPage.getDailyLimitNumber()).toEqual('0');
+  });
 });
--- a/src/wwwroot/controllers/ReportsCtrl.js
+++ b/src/wwwroot/controllers/ReportsCtrl.js
@@ -30,6 +30,9 @@
     $scope.getDailyLimit = function() {
       return $rootScope.accountLimits.daily && $rootScope.accountLimits.daily.limit;
     };
+    $scope.showDailyLimit = function() {
+      return !!$rootScope.accountLimits.daily;
+    };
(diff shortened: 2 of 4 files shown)


src/spec/reports_spec.js
```diff
@@ -211,7 +211,7 @@ describe('Reports page', () => {
     expect(reportsPage.isMaxRateDailyLimitDisplayed()).toEqual(true);
   });
 
-  it('should not show the daily limits for accounts that do not have limits', () => {
+  it('should show zero on the daily limits for accounts that do not have limits', () => {
```
andresmoschini: This fact is wrong: when the user _does not have limits_ he should not see the limits. He should see the limits **when there is a limit, even zero**.

src/wwwroot/controllers/ReportsCtrl.js
```diff
@@ -28,7 +28,8 @@
 
     $scope.maxRateDailyLimitBar = 50;
     $scope.getDailyLimit = function() {
-      return $rootScope.accountLimits.daily && $rootScope.accountLimits.daily.limit;
+      
+      return $rootScope.accountLimits.daily && $rootScope.accountLimits.daily.limit?$rootScope.accountLimits.daily.limit:0;
```
andresmoschini: It is wrong, it is not the same _limit of zero_ than _no limits_

src/spec/reports_spec.js
cristiangiagante: Oh ok, I' will update it. Thanks

src/wwwroot/controllers/ReportsCtrl.js
cristiangiagante: I' will update it. Thanks

cristiangiagante:
@andresmoschini @mvillaseco I updated the pr to show limits including limit zero and mantain the original logic.

src/wwwroot/controllers/ReportsCtrl.js
```diff
@@ -30,6 +30,9 @@
     $scope.getDailyLimit = function() {
       return $rootScope.accountLimits.daily && $rootScope.accountLimits.daily.limit;
     };
+    $scope.showDailyLimit = function() {
+      return $rootScope.accountLimits.daily;
```
andresmoschini: Is it possible having a response like this?

```json
{
    "daily": {
        "limit": null,
        "remaining": 0,
        "reset": "2017-09-16T00:00:00Z"
    }
}
```

Because in that case, it will not work.

Also, what do you think about convert the result to boolean to make it clear your intentions?

```js
return !!$rootScope.accountLimits.daily;
```

src/spec/reports_spec.js
```diff
@@ -240,4 +240,39 @@ describe('Reports page', () => {
...
+        });
+      }));
+    var reportsPage = new ReportsPage();
+
+    // Act
+    browser.get('/#/reports');
+
+    // Assert
+    expect(reportsPage.isMaxRateDailyLimitDisplayed()).toEqual(true);
+    expect(reportsPage.getDailyLimitNumber()).toEqual('0');
```
andresmoschini: Thanks for adding a new test!

src/wwwroot/controllers/ReportsCtrl.js
cristiangiagante: Hi @andresmoschini, about the first comment, is not possible to get a response like that.
This section of the SendRateLimit_To_LimitStatus mapper return null the entire daily property if the limit is null.
```c#
public void Fill(List<SendRateLimit> input, LimitStatus output)
        {
            output.monthly = GetLimitFromSendRateLimit(input.FirstOrDefault(r => r.LimitType == RateLimitType.Monthly));
            output.daily = GetLimitFromSendRateLimit(input.FirstOrDefault(r => r.LimitType == RateLimitType.Daily));
            output.hourly = GetLimitFromSendRateLimit(input.FirstOrDefault(r => r.LimitType == RateLimitType.Hourly));
        }

        private static Limit GetLimitFromSendRateLimit(SendRateLimit sendRateLimit)
        {
            if (sendRateLimit == null || !sendRateLimit.Limit.HasValue)
            {
                return null;
            }
           
            var remaining = (sendRateLimit.Limit.Value - sendRateLimit.CurrentCount);
            return new Limit
            {
                remaining = (remaining < 0) ? 0 : remaining,
                limit = sendRateLimit.Limit.Value,
                reset = sendRateLimit.ResetDate.Value.ToDateTimeOffset()
            };
        }
```

About the second comment, yes! I will do it. Thank you!

andresmoschini: You are right! also, the [schema](http://api.dopplerrelay.com/schemas/limit-status.json) indicates that _limit_ field is not nullable.

https://github.com/MakingSense/Doppler_Relay/blob/d31706fc2c9cf09ad4dd6b7ae6a335d1791b9b0c/Relay.Application.Api/wwwroot/schemas/limit-status.json#L24-L41

Thanks for the details.

cristiangiagante: Thanks for tip! :)

PR #235, opened 2018-04-27
[DR-1597] Remove Doppler relay word that it wasnt the same as spanish version

mvillaseco:
# Overview
 HotFix to remove "Doppler Relay" word in the english version to match the spanish. And also fix the problem.


andresmoschini:
👍

PR #236, opened 2018-05-08
[DR-1516] Set gire domain

cristiangiagante:
## Background
  
We need setup the real domain for the new brand
  
## Changes done
  
Update domain in utils script and reCaptcha account
  
## Pending to be done
  
~~Request TLS protocol~~
~~Set new domain in reCaptcha~~
  
## Notes
~~We are waiting confirmation about if the new domain will implement TLS and waiting for the reCaptcha account credentials from @dnoya~~

## Demo

image

### DIFF
--- a/src/wwwroot/services/utils.js
+++ b/src/wwwroot/services/utils.js
@@ -14,7 +14,7 @@
   ];
 
   var domainsData = [
-    { companyName: 'Gire', logoName: 'gire-logo-grey', domainRegex: /^gire$/, companyLink: "http://giresoluciones.com.ar" },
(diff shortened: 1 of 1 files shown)


andresmoschini:
Thanks @cristiangiagante !

Do not await confirmation about TLS, assume that they do not. If they need it, @dnoya will call us.

But do not close this task without configure reCaptcha correctly.

cristiangiagante:
reCaptcha is ready, @mvillaseco @andresmoschini if you are agree, I will to procede with the merge of the PR.

andresmoschini:
Nice! I think that it is ready to be merged. Please, deploy it to QA, touch your _host_ file and test it before close.

PR #239, opened 2018-05-09
[DR-1621] Add mime fonts to prevent missing default mime types in IIS

cristiangiagante:
## Background
  
Prevent missing default mime types on IIS servers
  
## Changes done
  
Add default mime types for ttf and eot font extensions
  
## Pending to be done
  
N/A
  
## Notes
  
RFC References: https://www.iana.org/assignments/media-types/media-types.xhtml

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types


andresmoschini:
:shipit:

PR #241, opened 2018-05-09
Remove custom mime font types to prevent collitions

cristiangiagante:
## Background
  
Possible collitions if IIS server have setted some mime font types
  
## Changes done
  
Remove custom mime types before override to prevent collitions with IIS

## Notes

Thanks @andresmoschini to alert me about this more earlier, I was wrong about the IIS behavior with overriding in mime types


andresmoschini:
:shipit:

PR #244, opened 2018-05-22
[DR-1628] Prototype image uploading

andresmoschini:
Hi @mvillaseco and @cristiangiagante, 

It adds a layer between DropZone and our backend and by the moment it has a double to avoid HTTP calls.

See related changes in MSEditor: https://github.com/MakingSense/MSEditor/pull/1259

### DIFF
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -23,7 +23,8 @@
       getPreviewImageWithPlay: getPreviewImageWithPlay,
       getSettings: getSettings,
       refreshMercadoShopsProductServiceToken: refreshMercadoShopsProductServiceToken,
-      getTiendaNubeProducts: getTiendaNubeProducts
+      getTiendaNubeProducts: getTiendaNubeProducts,
+      overrideUploadFiles: overrideUploadFiles
     };
 
     return service;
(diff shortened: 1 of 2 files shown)


andresmoschini:
By the way @cristiangiagante , why there are fake content in our MSEditor CDN?

imagen

Is it being generated in all builds? Is it something specific for Relay? (In that case I think that it should not be in MSEditor CDN, maybe you could upload in a Relay specific place, maybe something like `http://cdn.fromdoppler.com/relay-demo-content` or something like that.

cristiangiagante:
It was something specific to relay to check the method, the image was uploaded manually in the build #306. Ok I will modify it to check the http://cdn.fromdoppler.com/relay-demo-content path.

andresmoschini:
😴 💤 💤

src/wwwroot/template-editor/adapterService.js
```diff
@@ -23,7 +23,8 @@
       getPreviewImageWithPlay: getPreviewImageWithPlay,
       getSettings: getSettings,
       refreshMercadoShopsProductServiceToken: refreshMercadoShopsProductServiceToken,
-      getTiendaNubeProducts: getTiendaNubeProducts
+      getTiendaNubeProducts: getTiendaNubeProducts,
+      overrideUploadFiles: overrideUploadFiles
```
cristiangiagante: Could you help to me telling me where is being called the adapterService method overrideUploadFiles? I check (https://github.com/MakingSense/MSEditor/pull/1259) and I can't found it. I can see that dropzone.uploadFiles is being overrided but I can't found where dropzone.uploadFiles is called.

andresmoschini: Oh, it is called by dropzone, let me see...

Here it is: https://github.com/enyo/dropzone/blob/caf200c13fd3608dd6bed122926d5848927f55b4/dist/dropzone.js#L2272

src/wwwroot/template-editor/adapterService.js
```diff
@@ -415,6 +416,253 @@
...
+
+      var file, formData, handleError, headerName, headerValue, headers, i, input, inputName, inputType, key, method, option, progressObj, response, updateProgress, url, value, xhr, _i, _j, _k, _l, _len, _len1, _len2, _len3, _m, _ref, _ref1, _ref2, _ref3, _ref4, _ref5;
+
+      // DropZone's original code:
+      // method = resolveOption(this.options.method, files);
+      method = 'POST';
+
+      // DropZone's original code:
+      // url = resolveOption(this.options.url, files);
+      url = 'images';
```
cristiangiagante: This could be targeting a fake image on CDN. What do you think?

cristiangiagante:
I checked the code, it seems enough for our requirements, good job 👍

src/wwwroot/template-editor/adapterService.js
andresmoschini: This is the URL for upload the image, and also we are not uploading it thanks of our XMLHttpRequest double so I do not see any advantage on doing that.

cristiangiagante: Oh I see now, thanks andres :)

PR #250, opened 2018-05-31
[DR-1639] Get language from webapp local storage

cristiangiagante:
## Background
  
Integrate webapp language with mseditor
  
## Changes done
  
Add initializer in adapterService

Method to get language from localstorage 
  
## Pending to be done
  
N/A
  
## Notes

Locale files are stored locally.
 
## Demo
 
image

### DIFF
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -8,10 +8,14 @@
 
   adapterService.$inject = [
     '$http',
-    '$q'
+    '$q',
+    '$window'
   ];
 
-  function adapterService($http, $q) {
+  function adapterService(
+    $http,
+    $q,
+    $window) {
     var service = {
       getCampaign: getCampaign,
       campaignSaveChanges: campaignSaveChanges,
@@ -26,9 +30,21 @@
       getTiendaNubeProducts: getTiendaNubeProducts,
       overrideUploadFiles: overrideUploadFiles
     };
+    
+    var langKey = null;
+    init();
 
     return service;
 
+    function init() {
+      langKey = getPreferredLanguage();
+    };
+
+    function getPreferredLanguage() {
+      var key = $window.localStorage.getItem('lang');
+      return key || 'en';
+    };
+    
     /**
      * Return a preview placeholder image
      * @argument {Object} params - Configuration object
@@ -340,40 +356,12 @@
     function getSettings(idCampaign, idTemplate) {
       traceAdapterCall(getSettings, arguments);
       var msEditorSettings = {
-        language: 'es',
-        sharedSocialNetworks: [{
-          name: 'twitter',
-          url: 'http://twitter.com/share?related=fromdoppler',
-          idSocialNetwork: '0'
-        }, {
-          name: 'linkedin',
-          url: 'http://linkedin.com/',
-          idSocialNetwork: '1'
-        }, {
-          name: 'facebook',
-          url: 'http://facebook.com',
-          idSocialNetwork: '2'
-        }, {
-          name: 'googlemas',
-          url: 'http://google.com',
-          idSocialNetwork: '3'
-        }, {
-          name: 'pinterest',
-          url: 'http://pinterest.com',
-          idSocialNetwork: '4'
-        }, {
-          name: 'whatsapp',
-          url: 'http://whatsapp.com',
-          idSocialNetwork: '5'
-        }],
-        stores: [{
-          name: 'TiendaNube',
-          accessToken: '4b9883131f01fbde3f168dde5bca209b1db1fbb1',
-          storeId: '404554'
-        }],
-        redirectUrl: '/MSEditor/Editor?idCampaign=1&redirectedFromSummary=True',
-        rssCampaign: true,
-        rssShowPreview: true
+        language: langKey,
+        redirectUrl: "/",
+        sharedSocialNetworks: [],
+        stores: [],
+        rssCampaign: false,
+        rssShowPreview: false
       };
       return $q.resolve(msEditorSettings);
     }


src/wwwroot/template-editor/adapterService.js
```diff
@@ -8,10 +8,11 @@
 
   adapterService.$inject = [
     '$http',
-    '$q'
+    '$q',
+    '$window'
   ];
 
-  function adapterService($http, $q) {
+  function adapterService($http, $q, $window) {
```
andresmoschini: Is there not a _$localStorage_ or something like that?

src/wwwroot/template-editor/adapterService.js
```diff
@@ -340,7 +352,7 @@
     function getSettings(idCampaign, idTemplate) {
       traceAdapterCall(getSettings, arguments);
       var msEditorSettings = {
-        language: 'es',
+        language: langKey,
         sharedSocialNetworks: [{
```
andresmoschini: Since by the moment we have all that we need from this method, could you finish it?

I think that it is only pending to:

* removing sharedSocialNetworks
* removing stores
* make rss related properties _false_
* set _redirectUrl_ with a Doppler Relay URL

src/wwwroot/template-editor/adapterService.js
cristiangiagante: $window is the service used to access to browser features. I don't found any more in AngularJS api reference.

src/wwwroot/template-editor/adapterService.js
andresmoschini: You could also remove the line related to _traceAdapterCall_, but I am not sure, we can do that later.

cristiangiagante: Of course.

andresmoschini:
What do you think about renaming this PR as something like _Implement getSetting adapter method_?

src/wwwroot/template-editor/adapterService.js
andresmoschini: You are right, I saw an npm package for that, so, it seems that there is nothing out of the box. Keep it as it is right now. Thanks!

src/wwwroot/template-editor/adapterService.js
```diff
@@ -27,6 +28,17 @@
...
 
+    var langKey = null;
+    init();
+    
+    function init() {
+      langKey = getPreferredLanguage();
+    };
+
+    function getPreferredLanguage() {
+      return $window.localStorage.getItem('lang');
```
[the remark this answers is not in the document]
cristiangiagante: I guess is a nice to have but depends how we will handle the authentication state.
It is supposed that the language will be always setted by the webapp, then access directly to the mseditor has not sense taking in count if we will redirect to the webapp if the application is not authenticated.

Let me know your opinion. Thanks andres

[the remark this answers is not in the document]
andresmoschini: I take a quick look at WebApp and cannot understand how are we dealing with this situation :P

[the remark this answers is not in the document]
andresmoschini: >  then access directly to the mseditor has not sense taking in count if we will redirect to the webapp if the application is not authenticated.

Ok, but that behavior does not exist yet. Maybe in the future, we choose to show a popup, and, in what language?

src/wwwroot/template-editor/adapterService.js
andresmoschini: This is something totally opinable, so, feel free to take my advice or not.

Did you see that there is a line for each dependency in _adapterService.$inject_? I would link to use the same approach for parameters in this constructor:

```js
adapterService.$inject = [
  '$http',
  '$q'
  '$window'
];
 
function adapterService(
  $http, 
  $q, 
  $window) {
```

In my opinion, it is easier to read (when start to be a lot of them) and also it is friendly with the diffs.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -27,6 +28,18 @@
       overrideUploadFiles: overrideUploadFiles
     };
 
+    var langKey = null;
+    init();
+    
+    function init() {
```
andresmoschini: Could you declare these functions after _returning service instance_ in the same way that we are doing with the other functions?

src/wwwroot/template-editor/adapterService.js
```diff
@@ -340,40 +353,10 @@
     function getSettings(idCampaign, idTemplate) {
       traceAdapterCall(getSettings, arguments);
       var msEditorSettings = {
-        language: 'es',
-        sharedSocialNetworks: [{
```
andresmoschini: Will it work or do you need to return an empty array for _sharedSocialNetworks_?

src/wwwroot/template-editor/adapterService.js
```diff
@@ -340,40 +353,10 @@
...
-        }, {
-          name: 'pinterest',
-          url: 'http://pinterest.com',
-          idSocialNetwork: '4'
-        }, {
-          name: 'whatsapp',
-          url: 'http://whatsapp.com',
-          idSocialNetwork: '5'
-        }],
-        stores: [{
```
andresmoschini: Will it work or do you need to return an empty array for _stores_?

src/wwwroot/template-editor/adapterService.js
```diff
@@ -340,40 +353,10 @@
...
-          name: 'TiendaNube',
-          accessToken: '4b9883131f01fbde3f168dde5bca209b1db1fbb1',
-          storeId: '404554'
-        }],
-        redirectUrl: '/MSEditor/Editor?idCampaign=1&redirectedFromSummary=True',
-        rssCampaign: true,
-        rssShowPreview: true
+        language: langKey,
+        redirectUrl: '#/templates',
```
andresmoschini: I think that it will not work because you have to change the URL not only the hash

src/wwwroot/template-editor/adapterService.js
```diff
@@ -339,41 +352,22 @@
...
       traceAdapterCall(getSettings, arguments);
+      var url = "/#/"
+      if (idCampaign != 0)
+      {
+        url = "/template-editor/?idCampaign=" + idCampaign;
+      }
+      if (idTemplate != 0)
+      {
+        url = "/template-editor/?idTemplate=" + idTemplate;
+      }
```
cristiangiagante: This redirectUrl is not being used by back button, sorry, this is used to handle a specific match from response html in a httpInterceptor:
```js

    function response(response) {
      if (typeof response.data === 'string' && (response.data.indexOf('<title>Login - Doppler Email Marketing</title>') !== -1 ||
        response.data.indexOf('<title>Ingresa - Doppler Email Marketing</title>') !== -1)) {

        //We create a temporary internal message to emulate a 401 unauthorized message
        var error = {
          status: 401,
          message: 'Unauthorized'
        };

        showModalAndRedirect();

        return $q.reject(error);
      }

      return response;
    }
```

this interceptor calls to:
```js
    function showModalAndRedirect() {
     var settingsPromise;
     var changesManager;

     var modalOptions = {
        title: $translate.instant('errors.server_error'),
        message: $translate.instant('errors.saving_changes'),
        buttons: [{
          label: $translate.instant('accept'),
          class: 'button--primary'
        }]
      };

      return modalService.showModal(modalOptions).then(function() {
        settingsPromise = $injector.get('settingsService').getSettings();
        changesManager = $injector.get('changesManager');

        return settingsPromise.then(function(settings) {
          changesManager.setUnsavedChangesCampaign(false);
          $window.location.href = settings.redirectUrl;
        });
      });
    }
```

Here you will found the redirection using this json attribute:
`$window.location.href = settings.redirectUrl`

This is the only one place where is used. I checked the doppler backend and this attribute is being setted depending the value of idCampaigns and idTemplates, but the MSEditor app is handling this behavior too.

MSEditor catch the value of idCampaign or idTemplates:
```js
if(isNaN(idCampaign) && isNaN(idTemplate) || idCampaign && idTemplate) {
              return $q.reject('Url should have an idCampaign or an idTemplate as in: "/?idCampaign=1" or "/?idTemplate=1"');
            }
```

Then, the behavior on the interceptor is never used. I think maybe this can be a security patch or something like that.
I found in google the <title> being listened in the interceptor and match with **https://cm.fromdoppler.com/**

I really think that we don't need this setting, but by the moment I will handle locally with the same logic that the backend in doppler used.

Doppler backend getSettings method:
```c#
if ((idCampaign == 0) && (idTemplate == 0))
{
	dtoEditorSetting.RedirectURL = "/SignIn";
}
else
{
	if (idCampaign != 0)
	{
                // More code here
		dtoEditorSetting.RedirectURL = "/MSEditor/Editor?idCampaign=" + idCampaign.ToString();
	}
	if (idTemplate != 0)
	{
                // More code here
		dtoEditorSetting.RedirectURL = "/MSEditor/Editor?idTemplate=" + idTemplate.ToString();
	}
}
```

Sorry for the big comment, let me know your thoughts.

src/wwwroot/template-editor/adapterService.js
cristiangiagante: We will return a empty array of stores and we will update MSEditor logic to handle sharedSocialNetworks when is empty.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -26,9 +30,30 @@
...
       overrideUploadFiles: overrideUploadFiles
     };
+    
+    var langKey = null;
+    init();
 
     return service;
 
+    function init() {
+      console.log("test");
```
andresmoschini: Remember to remove this line.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -26,9 +30,30 @@
...
+    };
+
+    /**
+     * Return a language identifier value stored previously in the local storage of the browser
+     * @example
+     * ```js
+     * // Return 'en' or 'es'
+     * var langKey = getPreferredLanguage();
+     * ```
+     */
```
andresmoschini: In my opinion, this documentation is not necessary, this code in enough clear by his own, and also it is a private function.

src/wwwroot/template-editor/adapterService.js
andresmoschini: I think that I am understanding. It seems that backend work is not necessary.
Lets talk about it when you have time.

PR #251, opened 2018-06-01
[DR-1640] Use auth token in mseditor

cristiangiagante:
## Background
  
We need get the auth token from webapp in mseditor app to send to backend
  
## Changes done
  
Refactor on webapp auth service to attend new loginSession wrapper for token

MSeditor method to get loginSession with needed info for backend api calls

Fix for data lack in tests (missing relay_temporal_token property for temporary sessions)
  
## Pending to be done
  
N/A
  
## Notes
  
This subtask scope's only cover to get the auth token in mseditor. Handle expiration date and errors in MSEditor will be in another task.
 
## Demo

image

image


src/wwwroot/template-editor/adapterService.js
```diff
@@ -38,8 +41,32 @@
...
+      var token = $window.localStorage.getItem('jwtToken');
+      if (!token) {
+        logOut();
+      } else if (token != encodedToken) {
+        encodedToken = token;
+        return jwtHelper.decodeToken(encodedToken);
+      }
+    }
+
+    function logOut() {
```
andresmoschini: What do you think about delegating this kind of things to WebApp? I mean, redirecting or something like that (in order of keeping it simple)

src/wwwroot/template-editor/adapterService.js
```diff
@@ -38,8 +41,32 @@
 
     function init() {
       langKey = getPreferredLanguage();
+      decodedToken = getToken();
```
andresmoschini: If I am not wrong, you do not need to decode the token to use it. So, maybe you do not need to include _jwtHelper_.

I know that you need some more information from the token, but maybe we could prepare it using the WebApp, for example storing the accountname and userid in the local storage.

andresmoschini: Or passing it as parameter

src/wwwroot/template-editor/adapterService.js
cristiangiagante: Good idea 🔢

src/wwwroot/template-editor/adapterService.js
cristiangiagante: I guess the reason to use the token is to not allow pass this information in plain text mode by the network, maybe we can store the extra data needed in localstorage but I guess we need to think in the security concerns.

andresmoschini: The token itself is the secret, so, it is already plain text.

_UserId_ and _accountname_ are not secrets.

andresmoschini: What is your security concern?

cristiangiagante: My concerns was if UserId and accountname can be stored in plain text, but if it's not a problem I can update that to avoid inject the angular-jwt module. Thanks andres

cristiangiagante: Finally we will use a login json object in the webapp to wrapping jwtToken and include the new properties for mseditor

src/wwwroot/services/auth.js
```diff
@@ -37,25 +37,41 @@
...
+      if (encodedToken) {
+        decodeToken();
+        // verify expiration
+      }
+    }
+
+    function getLoginSession(storageName) {
+      var login = angular.fromJson($window.localStorage.getItem(storageName));
+      return login;
+    }
```
andresmoschini: I am not sure if this function is necessary.

src/wwwroot/services/auth.js
```diff
@@ -37,25 +37,41 @@
       isUrlAllowed: isUrlAllowed,
       getDefaultUrl: getDefaultUrl
     };
-
+    var loginSession = null;
     var decodedToken = null;
     var encodedToken = null;
```
andresmoschini: Do you think that it is possible to remove _decodedToken_ or at least _encodedToken_ and use _loginSession_ in place?

src/wwwroot/services/auth.js
```diff
@@ -37,25 +37,41 @@
...
+
     // Save the token in the local storage (globally) or in a temporal variable (local)
     function saveToken(token, isTemporal) {
       encodedToken = token;
       decodeToken();
       temporarilyAuthed = !!isTemporal;
       if (!temporarilyAuthed) {
-        $window.localStorage.setItem('jwtToken', token);
+        loginSession = {
+          jwtToken: token,
```
andresmoschini: What do you think about renaming it as _token_?

cristiangiagante: Ok, I think is fine.

src/wwwroot/services/auth.js
```diff
@@ -37,25 +37,41 @@
...
     return authService;
 
+    function init() {
+      loginSession = getLoginSession('relayLogin');
+      encodedToken = loginSession? loginSession.jwtToken : null;
+      if (encodedToken) {
+        decodeToken();
+        // verify expiration
+      }
+    }
```
andresmoschini: You are not keeping backward compatibility with old _jwtToken_ local storage entry, or am I missing something?

src/wwwroot/services/auth.js
```diff
@@ -37,25 +37,41 @@
...
-    if (encodedToken) {
-      decodeToken();
-      // verify expiration
-    }
+    init();
     return authService;
 
+    function init() {
+      loginSession = getLoginSession('relayLogin');
+      encodedToken = loginSession? loginSession.jwtToken : null;
```
andresmoschini: Please, start using space before ternary operator: 

```
encodedToken = loginSession ? loginSession.jwtToken : null;
                           ˆ
```

src/wwwroot/services/auth.js
```diff
@@ -168,7 +184,7 @@
     function ensureToken() {
       // It is to avoid have different users in UI and AJAX requests
       // TODO: improve it because when token is removed, the request is done anyway, maybe it is possible to cancel request when there is no token
-      var token = $window.localStorage.getItem('jwtToken');
+      var token = loginSession? loginSession.jwtToken : null;
```
andresmoschini: In the previous implementation we where using local storage value, but now you are using _in memory_ value, also, there is a TODO note that maybe should be taken into account.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -32,18 +32,59 @@
...
+
+    function getAccountId() {
+      if (isTokenExpired()) {
+        logOut();
+      }
+      return loginSession.accountId;
+    }
+
+    function getAccountName() {
+      if (isTokenExpired()) {
```
andresmoschini: Please, in order to simplify remove all references to something not implemented.

src/wwwroot/services/auth.js
andresmoschini: I think that it is possible, maybe you should replace _decodedToken_ by _loginSession_

src/wwwroot/services/auth.js
cristiangiagante: Yes, you are right I will check that.

src/wwwroot/services/auth.js
cristiangiagante: I will remove it to call directly angular.fromJson method and covert the local storage.

src/wwwroot/services/auth.js
cristiangiagante: I don't really sure about that, we need to define where will be stored the permission account info.

src/wwwroot/services/auth.js
cristiangiagante: Yes it is. encodedToken is being loaded from local storage, then I read loginSession from local storage through getLoginSession('relayLogin') and passing the token to encodedToken.
*I will remove getLoginSession to do this more easy to understand and clean.

cristiangiagante:
@mvillaseco could you let me know your comments about this changes, thanks!

cristiangiagante:
@andresmoschini I think it's ready. @mvillaseco I tested it and is working as expected.

andresmoschini:
Hi @cristiangiagante, I am a little delayed, but I want to review this PR before merging.

src/wwwroot/services/auth.js
```diff
@@ -37,31 +37,56 @@
...
+      var encodedToken = $window.localStorage.getItem('jwtToken');
+      if (encodedToken) {
+        try {
+          loginSession = decodeLoginSession(encodedToken);
+        } catch {
+          logOut();
+          return;
+        }
+      }
+      // To ensure having relayLogin item in legacy sessions
```
andresmoschini: If I am not wrong, this code should be only executed if _encodedToken_ has value.

src/wwwroot/services/auth.js
```diff
@@ -37,31 +37,56 @@
...
+      var storedSession = $window.localStorage.getItem('relayLogin');
+      if (encodedToken && !storedSession) {
```
andresmoschini: > If I am not wrong, this code should be only executed if encodedToken has value.
Oh, you are verifying that here, what do you think about moving it inside _if block_ (`if (encodedToken) {`)?

src/wwwroot/services/auth.js
```diff
@@ -157,73 +182,57 @@
...
+        return loginSession.token
       }
+      ensureToken();
+      return loginSession && loginSession.token;
     }
-
+    
     function ensureToken() {
-      // It is to avoid have different users in UI and AJAX requests
-      // TODO: improve it because when token is removed, the request is done anyway, maybe it is possible to cancel request when there is no token
```
andresmoschini: What do you think about resurrecting these comments? 

I think that the _TODO note_ is still relevant, and we could clarify a little more the explanation in the first line, what do you think about:

```
// It is to allow apply logout in all open tabs and also to avoid having different users in UI and AJAX requests.
```

src/wwwroot/template-editor/adapterService.js
```diff
@@ -32,19 +32,43 @@
...
+    }
+
     function getPreferredLanguage() {
       var key = $window.localStorage.getItem('lang');
       return key || 'en';
     };
-    
+
+    function logOut() {
+      $window.location = "/";
```
andresmoschini: Do you think that it could be useful cleaning session variables before redirect?

src/wwwroot/services/auth.js
cristiangiagante: Ok! 👍

andresmoschini:
Great work @cristiangiagante !

I think that we will also need something similar to _ensureToken_ inside the adapter, but we can postpone that for the next task/pr

src/wwwroot/services/auth.js
cristiangiagante: I will add them. thx

src/wwwroot/template-editor/adapterService.js
cristiangiagante: Well I think that is not necessary due to when we change the url location, in memory variables from the service will be cleaned and localstorage will be managed by the webapp.

PR #252, opened 2018-06-06
[DR-1644] Call to backend and handle error for getCampaign

cristiangiagante:
## Background
  
Create a js bundle with version  number to mseditor's local dependencies
  
## Changes done
  
Update getCampaign method to send request

Handle errors

Log errors in console

## Pending to be done
  
N/A
  
## Notes

RELAY_CONFIG service was replaced for a mocked one until the environment config integration is ready

## Demo
Error:
image

### DIFF
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -9,13 +9,15 @@
   adapterService.$inject = [
     '$http',
     '$q',
-    '$window'
+    '$window',
+    'RELAY_CONFIG'
   ];
 
   function adapterService(
     $http,
     $q,
-    $window) {
+    $window,
+    RELAY_CONFIG) {
     var service = {
       getCampaign: getCampaign,
       campaignSaveChanges: campaignSaveChanges,
@@ -354,15 +356,25 @@
      */
     function getCampaign(id, useEditorAsTemplate) {
       traceAdapterCall(getCampaign, arguments);
-      var campaign = {
-        id: id,
-        type: 'campaign',
-        name: 'Campaign ' + id,
-        attributes: {},
-        innerHTML: '',
-        children: []
-      };
-      return $q.resolve({ data: campaign });
+      return $http.get(`${RELAY_CONFIG.baseUrl}/accounts/${loginSession.accountName}/template/${id}`, {
+        headers: {
+          'Authorization': 'Bearer '+ apiToken
+        }
+      }).then(function(response) {
+        var campaign = {
+          id: id,
+          type: 'campaign',
+          name: 'Campaign ' + id,
+          attributes: {},
+          innerHTML: '',
+          children: []
+        };
+        return { data: campaign };
+      },
+      function(error) {
+        alert(error.data && error.data.detail || "Unexpected exception");
+        $window.location = "/#/templates"
+      });
     }
 
     /**
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -9,15 +9,13 @@
   adapterService.$inject = [
     '$http',
     '$q',
-    '$window',
-    'RELAY_CONFIG'
+    '$window'
   ];
 
   function adapterService(
     $http,
     $q,
-    $window,
-    RELAY_CONFIG) {
+    $window) {
     var service = {
       getCampaign: getCampaign,
       campaignSaveChanges: campaignSaveChanges,
@@ -36,6 +34,9 @@
     var langKey = null;
     var loginSession = null;
     var apiToken = null;
+    var RELAY_CONFIG = {
+      baseUrl: 'http://localhost:34751'
+    };
     init();
 
     return service;
@@ -356,7 +357,8 @@
      */
     function getCampaign(id, useEditorAsTemplate) {
       traceAdapterCall(getCampaign, arguments);
-      return $http.get(`${RELAY_CONFIG.baseUrl}/accounts/${loginSession.accountName}/template/${id}`, {
+      // TODO This is a temporary url reference to the current backend syntax call
+      return $http.get(`${RELAY_CONFIG.baseUrl}/accounts/${loginSession.accountId}/templates/${id}`, {
         headers: {
           'Authorization': 'Bearer '+ apiToken
         }
@@ -372,11 +374,11 @@
         return { data: campaign };
       },
       function(error) {
-        alert(error.data && error.data.detail || "Unexpected exception");
-        $window.location = "/#/templates"
+        var errorDetail = error.data && error.data.detail || "Unexpected error";
+        console.log(errorDetail);
+        return $q.reject(error);
       });
     }
-
     /**
      * Return settings from json to configure language, 3rd service tokens, and url links
      * @argument {number} idCampaign - The campaign identifier


cristiangiagante:
@andresmoschini could you check the changes to handle environment dependencies in mseditor? thanks

src/gulpfile.js
```diff
@@ -255,6 +255,25 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
   .pipe(gulp.dest(paths.build));
 });
 
+gulp.task('build-mseditor', ['template-editor'], function () {
+  return gulp.src([
+    paths.app + '/template-editor/index.html'
+  ])
+  .pipe(gulpInject(gulp.src(paths.app + '/env/template-editor/' + process.env.NODE_ENV + '.js'), {
+    transform: function(filePath, file) {
+      return '<script>' + file.contents.toString() + '</script>';
```
andresmoschini: As I understand, it is injecting a module as inline script.

What do you think about bundle all .js files in a single one in place of doing it?

src/wwwroot/env/template-editor/INT.js
```diff
@@ -0,0 +1,5 @@
+﻿angular.module('mseditor').constant('RELAY_CONFIG', {
```
andresmoschini: Is it possible using the same configuration files for WebApp and templates editor?

src/wwwroot/template-editor/adapterService.js
```diff
@@ -354,15 +356,27 @@
...
+          type: 'campaign',
+          name: 'Campaign ' + id,
+          attributes: {},
+          innerHTML: '',
+          children: []
+        };
+        return { data: campaign };
+      },
+      function(error) {
+        $window.location = "/#/templates"
```
andresmoschini: I am not sure about it. What do you think about a popup or an alert and then redirect or refresh?

src/gulpfile.js
```diff
@@ -255,6 +255,25 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+    }
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
+});
+
+gulp.task('template-editor', function () {
+  return gulp.src([
+      paths.app + '/template-editor/adapterService.js'
```
andresmoschini: Oh, we forgot it, but we should include revision number in this js, in another case we could cache unwanted things.

Maybe bundle all js files could help on that.

cristiangiagante: Yes, the bundle approach will prevent the cache

src/wwwroot/template-editor/adapterService.js
cristiangiagante: Ok, I will add an alert describing the error.

src/gulpfile.js
cristiangiagante: I will check it, we can add the serviceAdapter.js and the environment js  file defining the backend configuration.

src/wwwroot/env/template-editor/INT.js
cristiangiagante: I don't think so but, maybe we can move the environment related files to a env folder inseide template editor due to template editor is separated of webapp and handle the bundle from the gulpfile mergin js files inside template editor folder.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -354,15 +356,28 @@
...
-      var campaign = {
-        id: id,
-        type: 'campaign',
-        name: 'Campaign ' + id,
-        attributes: {},
-        innerHTML: '',
-        children: []
-      };
-      return $q.resolve({ data: campaign });
+      return $http.get(`${RELAY_CONFIG.baseUrl}/accounts/${loginSession.accountId}/template/${id}`, {
```
andresmoschini: It could be great using accountName in place of accountId. Probably you will need to use a URL template engine in place of simply concatenating strings, in order to take into account URL encoding and other weird cases.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -354,15 +356,28 @@
...
+        params: {
+          useAsTemplate: useEditorAsTemplate
```
andresmoschini: I think the we could ignore this parameter.

cristiangiagante: 👍

src/wwwroot/template-editor/adapterService.js
cristiangiagante: Do you have a favourite URL template engine  to use for this case? I was checking this:
https://github.com/bramstein/url-template

andresmoschini: In WebApp we are simply concatenating the sections, so, we do not need to do in better here. Keep the concatenation.

cristiangiagante:
@andresmoschini Yesterday I find in mseditor something to show modal windows and I found the modalService, then I implement it. Please let me know your thougths

src/wwwroot/template-editor/adapterService.js
```diff
@@ -354,17 +375,28 @@
...
-      var campaign = {
-        id: id,
-        type: 'campaign',
-        name: 'Campaign ' + id,
-        attributes: {},
-        innerHTML: '',
-        children: []
-      };
-      return $q.resolve({ data: campaign });
+      return $http.get(`${RELAY_CONFIG.baseUrl}/accounts/${loginSession.accountName}/template/${id}`, {
```
andresmoschini: Hehe, it is my fault, yesterday I asked you to use _accountName_ in place of _accountId_, but all template methods for the moment are using _accountId_.

Anyway, we will call another URL, for example, something like `/accounts/{accountName}/template/{id}/content`, so, what do you think about to go back to _accontId_ and add a _TODO_ comment indicating that it is a temporal URL?

src/wwwroot/template-editor/index.html
```diff
@@ -15,6 +15,7 @@
 
     <ng-view></ng-view>
     <mseditor-loader class="preloader">
+      <mseditor-modal></mseditor-modal>
```
andresmoschini: Is it fine adding this element inside the loader? cc: @mvillaseco

src/wwwroot/template-editor/adapterService.js
```diff
@@ -354,17 +375,28 @@
...
+          attributes: {},
+          innerHTML: '',
+          children: []
+        };
+        return { data: campaign };
+      },
+      function(error) {
+        var errorDetail = error.data && error.data.detail || "Unexpected error";
+        showModalAndRedirect("/#/templates", errorDetail);
+        return $q.reject();
```
andresmoschini: I am not sure, but in my opinion, this behavior is not correct:

```js
function(error) {
  // . . .
  showModalAndRedirect("/#/templates", errorDetail);
  return $q.reject();
}
```

It is returning the error (`return $q.reject()`), to continue the normal error flow, in parallel of showing the dialog (`showModalAndRedirect("/#/templates", errorDetail);`). 

In my opinion, we should wait for the user closing the dialog before continue with the error flow, the code sould be something like:

```js
function(error) {
  // . . .
  return showModalAndRedirect("/#/templates", errorDetail).then(function() {
    return $q.reject(error);
  });
}
```

Notice that you are also losing the original error, I think that the right way is returning `return $q.reject(error);`, but please research for it.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -69,6 +76,20 @@
       $window.location = "/";
     }
 
+    function showModalAndRedirect(url, errorDetail) {
+      var modalOptions = {
+        title: $translate.instant('errors.server_error'),
+        message: errorDetail ,
+        buttons: [{
+          label: $translate.instant('accept'),
```
andresmoschini: What do you think about being more explicit about the behavior in the button label? for example: `Redirect to template list`. 

Maybe it is not so simple, we should think about how to inject our own custom texts. So, if it is not so easy, we could do it later.

src/wwwroot/template-editor/adapterService.js
cristiangiagante: It good for me, I will revert it.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -354,17 +375,28 @@
...
+        return { data: campaign };
+      },
+      function(error) {
+        var errorDetail = error.data && error.data.detail || "Unexpected error";
+        showModalAndRedirect("/#/templates", errorDetail);
+        return $q.reject();
+      });
     }
-
+
```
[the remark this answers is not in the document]
cristiangiagante: Ok

src/wwwroot/template-editor/adapterService.js
cristiangiagante: I will do it, thanks andres

src/wwwroot/template-editor/adapterService.js
cristiangiagante: I guess there is not an easy and cleanest way to do that. I asked in the $translate github repo for tips about it. If they answer me with a nice approach I will do that in another task. You are agree @andresmoschini  to left this for later until I have better information?.

andresmoschini: > I asked in the $translate github repo for tips about it

In that case, it is a good idea adding a link here

>  If they answer me with a nice approach I will do that in another task. You are agree @andresmoschini to left this for later until I have better information?.

As I said before, it is something that we will need, but for the moment "accept" is fine. So, do not worry about it to finish this task.

andresmoschini: It is only for future reference, do not apply it now. 

Googling for another thing I casually found this: http://embed.plnkr.co/v1QsiEx4E11Z7Ba3irOp/

It seems to be related to what we will need.

src/wwwroot/template-editor/index.html
andresmoschini: I have asked @mvillaseco, we agree in that hierarchically it should not be inside `mseditor-loader`, is there any reason to put it there?

If not, maybe between `<ng-view></ng-view>` and `<mseditor-loader class="preloader">` could be a good place.

andresmoschini:
:shipit: ! There is only a minor thing to do related to HTML hierarchy.

src/wwwroot/template-editor/index.html
cristiangiagante: The reason was that the mseditor-modal directive is covering the entire page then when async calls are made from backend any modal window is being showed behind of this element.

andresmoschini: @mvillaseco what do you think? Is there a more elegant solution?

If it is the only way, please add a comment next to it to help to understand it better and avoid unwanted changes.

---

Also, please think about it, because it can orient some code refactoring in the near future, and an improvement in MSEditor code base: We are showing an error modal in a really lower layer (adapter is the lowest one), it is a smell, it would be great having something to catch expected and unexpected errors in an upper layer. I mean, showing the error and redirect should not be a responsibility of our adapter, it is a responsibility of MSEditor.

It is not important to apply any refactoring right now but take into account, because probably we have to deal with similar cases soon.

cristiangiagante: We decided to left for later this implementation and refactor, and mantain the original behavior of mseditor

cristiangiagante:
I will merge it at the end of the day.

PR #253, opened 2018-06-07
Fix uglify issue in build process

cristiangiagante:
## Background
  
Fix issue in gulp build process in uglify library
  
## Changes done
  
Add error paremeter in catch clausule
  
## Pending to be done
  
N/A
  
## Notes
  
N/A
 
## Demo
 
image

### DIFF
--- a/src/wwwroot/services/auth.js
+++ b/src/wwwroot/services/auth.js
@@ -50,7 +50,7 @@
           // To ensure having relayLogin item in legacy sessions
           var storedSession = $window.localStorage.getItem('relayLogin');
           !storedSession && saveStoredSession(loginSession);
-        } catch {
+        } catch (error) {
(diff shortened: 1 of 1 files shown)


src/wwwroot/services/auth.js
```diff
@@ -50,7 +50,7 @@
           // To ensure having relayLogin item in legacy sessions
           var storedSession = $window.localStorage.getItem('relayLogin');
           !storedSession && saveStoredSession(loginSession);
-        } catch {
+        } catch(error) {
```
andresmoschini: Could you add this space in both cases?

```
        } catch (error) {
               ^
```

andresmoschini:
Good catch @cristiangiagante ! Thanks!

PR #271, opened 2018-06-12
[DR-1644] Create bundle js to template editor

cristiangiagante:
## Background
  
We need to bundle the local js files used in the template editor
  
## Changes done
  
Created new tasks to create js bundle

Inject bundle name in index.html
  
## Pending to be done
  
N/A
  
## Notes
  
The bundle is generated by environment files from wwwroot/env

### DIFF
--- a/src/gulpfile.js
+++ b/src/gulpfile.js
@@ -255,6 +255,31 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
   .pipe(gulp.dest(paths.build));
 });
 
+gulp.task('build-mseditor', ['build-scripts-template-editor', 'add-revision-numbers'], function () {
+  var sources = gulp.src([
+    paths.build + '/scripts/relay-editor.*.js'
+  ]);
+  return gulp.src([
+    paths.app + '/template-editor/index.html'
+  ])
+  .pipe(gulpInject(sources, {
+    addRootSlash: false, // ensures proper relative paths removing the root slash
+    ignorePath: paths.build // ensures proper relative paths removing the absolute path
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
+});
+
+gulp.task('build-scripts-template-editor', function () {
+  return gulp.src([
+    paths.app + '/template-editor/*.js'
+  ])
+  .pipe(concat('relay-editor.js'))
+  .pipe(uglify({
+    mangle: false
+  }))
+  .pipe(gulp.dest(paths.tmpPrebuild + '/scripts'));
+});
+
 gulp.task('build-partials', function () {
   // TODO: minify partials. It is important for cache/revision
   // TODO: take into account revision here
@@ -310,7 +335,7 @@ gulp.task('fonts', function () {
   .pipe(gulp.dest(paths.build + '/fonts'));
 });
 
-gulp.task('add-revision-numbers', ['build-scripts', 'build-styles', 'locales', 'svg-sprite', 'build-partials', 'fonts'], function () {
+gulp.task('add-revision-numbers', ['build-scripts-template-editor', 'build-scripts', 'build-styles', 'locales', 'svg-sprite', 'build-partials', 'fonts'], function () {
   var revAll = new RevAll();
   return gulp.src([
     paths.tmpPrebuild + '/**'
@@ -439,6 +464,7 @@ gulp.task('test', [
  */
 gulp.task('build', ['clean'], function () {
   gulp.start(
+    'build-mseditor',
     'build-html',
     'web.config',
     'hcheck.png'
--- a/src/wwwroot/template-editor/index.html
+++ b/src/wwwroot/template-editor/index.html
@@ -26,7 +26,10 @@
     </mseditor-loader>
   <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build349/scripts/lib.min.js"></script>
   <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build349/scripts/app.min.js"></script>
+  <!-- Inject the environment module dependency in build process for template editor-->
+  <!-- inject:js -->
   <script src="./adapterService.js"></script>
+  <!-- endinject -->
 </body>
 
 </html>


src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+gulp.task('build-mseditor', ['template-editor', 'add-revision-numbers'], function () {
+  var sources = gulp.src([
+    paths.build + '/template-editor/relayEditor.*.js'
+  ]);
+  return gulp.src([
+    paths.app + '/template-editor/index.html'
+  ])
+  .pipe(gulpInject(sources, {
+    addRootSlash: false, // ensures proper relative paths
+    ignorePath: paths.build + '/template-editor' // ensures proper relative paths
```
andresmoschini: I do not understand your comments `// ensures proper relative paths`, could you give me some examples of how `addRootSlash` and `ignorePath` affect the _injection_?

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
+});
+
+gulp.task('template-editor', function () {
+  return gulp.src([
+    paths.app + '/template-editor/adapterService.js',
```
andresmoschini: What do you think about including all files in the folder `paths.app + '/template-editor/'`?

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+    paths.app + '/env/' + process.env.NODE_ENV + '.js'
+  ])
+  .pipe(concat('relayEditor.js'))
+  .pipe(gulp.dest(paths.tmpPrebuild + '/template-editor'));
```
andresmoschini: If I understand it fine, it generates a file called `relayEditor.js` in `tmpPrebuild` folder, **who** read that file?

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+    paths.app + '/template-editor/index.html'
+  ])
+  .pipe(gulpInject(sources, {
+    addRootSlash: false, // ensures proper relative paths
+    ignorePath: paths.build + '/template-editor' // ensures proper relative paths
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
+});
+
+gulp.task('template-editor', function () {
```
andresmoschini: What do you think about calling this task `build-scripts-template-editor`? If I am not wrong it would be consistent with a similar task:

https://github.com/cristiangiagante/relay-webapp/blob/f92062907fd4ae94ab4c5466d637e4d373a6e807/src/gulpfile.js#L107

src/gulpfile.js
andresmoschini: Did you avoid _uglify_ for any reason?

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
   .pipe(gulp.dest(paths.build));
 });
 
+gulp.task('build-mseditor', ['template-editor', 'add-revision-numbers'], function () {
```
andresmoschini: If I am not wrong, as it is right now, `template-editor` is a facto dependency of `add-revision-numbers`.

What do you think about adding as dependency explicitly? or, another alternative is apply the revision numbers independently of the other scripts.

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+    addRootSlash: false, // ensures proper relative paths
+    ignorePath: paths.build + '/template-editor' // ensures proper relative paths
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
+});
+
+gulp.task('template-editor', function () {
+  return gulp.src([
+    paths.app + '/template-editor/adapterService.js',
+    paths.app + '/env/' + process.env.NODE_ENV + '.js'
```
andresmoschini: Does it work? or throw a kind of error because the module does not exist as you said to me before?

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+    paths.build + '/template-editor/relayEditor.*.js'
+  ]);
+  return gulp.src([
+    paths.app + '/template-editor/index.html'
+  ])
+  .pipe(gulpInject(sources, {
+    addRootSlash: false, // ensures proper relative paths
+    ignorePath: paths.build + '/template-editor' // ensures proper relative paths
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
```
andresmoschini: What do you think about keeping it in the same folder than the rest of the js files?

src/gulpfile.js
```diff
@@ -255,6 +255,29 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
   .pipe(gulp.dest(paths.build));
 });
 
+gulp.task('build-mseditor', ['template-editor', 'add-revision-numbers'], function () {
+  var sources = gulp.src([
+    paths.build + '/template-editor/relayEditor.*.js'
```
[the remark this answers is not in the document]
cristiangiagante: I was thinking in be specific but we can do it too. Do you think it is more clean?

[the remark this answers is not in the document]
andresmoschini: Maybe you could take the same approach that we are using for webapp, probably is the current one.

src/gulpfile.js
cristiangiagante: Of course, here you have some examples:

Suppose this full path: "http://.../example/myFile.js"

ignorePath: '/example' will filter the path with this result: /myFile.js

addRootSlash: will add (if not exist) or remove the first / for paths like "/myFile.js" with this results:
if it's true "myFile.js", and for false: "/myFile.js"

src/gulpfile.js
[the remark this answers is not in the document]
cristiangiagante: 👍

src/gulpfile.js
cristiangiagante: Ok 👍

src/gulpfile.js
cristiangiagante: It is loaded by add-revision-numbers to generate revision numbers

cristiangiagante: Not really, I only forgot it. Sorry, I will add it.

src/gulpfile.js
cristiangiagante: I like the first option, this task it is being used to do the revision on all the build , I like continue using it than create a parallel one, thanks!

src/gulpfile.js
cristiangiagante: It is throwing this error:
image

src/gulpfile.js
cristiangiagante: You are talking about the bundle file or the index.html? If you refer to the bundle js file, yes, we can move it to the script folder in the build, I will do it.

src/gulpfile.js
andresmoschini: In that case, I prefer removing this line or fix the error before the merge.

src/gulpfile.js
andresmoschini: Oh, you are right, this line of code refers to the HTML.

The HTML is fine in the folder `template-editor`, I was talking about the .js file.

src/gulpfile.js
andresmoschini: Thanks your explanation was very useful, Now I can imagine how it works.

src/gulpfile.js
cristiangiagante: Perfect. I will merge it, after to take cover the dependency task in DR-1649

src/gulpfile.js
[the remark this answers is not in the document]
cristiangiagante: I will mantain this approach but targeting to scripts folder, we need to do reference to this specific file to inject in the template-editor/index.html, if I use ** all the js files will be injected on the template editor index file and we don't want that.

cristiangiagante:
@andresmoschini I update the code to follow our webapp guidelines

src/gulpfile.js
```diff
@@ -255,6 +255,31 @@ gulp.task('build-html', ['add-revision-numbers'], function () {
...
+    ignorePath: paths.build // ensures proper relative paths removing the absolute path
+  }))
+  .pipe(gulp.dest(paths.build + '/template-editor'));
+});
+
+gulp.task('build-scripts-template-editor', function () {
+  return gulp.src([
+    paths.app + '/template-editor/*.js'
+  ])
+  .pipe(concat('relayEditor.js'))
```
andresmoschini: What is this file? why is it not in _template-editor_ folder?

andresmoschini: Ohh, is it the destination name?

andresmoschini: In that case could you use a lower case name?

andresmoschini:
:shipit: ! I have only an observation about _relayEditor.js_ but if I am not wrong it is not so important.

src/gulpfile.js
cristiangiagante: Yes it is the destination name, I will change for a lower case.

PR #274, opened 2018-06-15
[DR-1653] Update MSEditor version to load languages from CDN

andresmoschini:
(I will update the version number after merging https://github.com/MakingSense/MSEditor/pull/1268)


cristiangiagante:
Ok!

PR #276, opened 2018-06-15
Improve Travis CI and add Docker scripts

andresmoschini:
Hi @mvillaseco and @cristiangiagante,

With this change, Travis CI will run the build, and in that way, we will detect if something goes wrong with the build process.

I have also added our docker scripts, to allow me (and you too) to run and build it without installing the dependencies. I added _build_n_publish_, I know that by the moment we are not publishing the WebApp to our CDN, but I think that we could do it soon (1 year? :P).

Finally, I applied some cleaning in Travis CI process.

Could you kindly review?

**UPDATE:** I have fixed some other issues:

* QA Deployment process was not working (for about a month!)
* Reference to template editor .js file was wrong


cristiangiagante:
I'm agree with this changes

PR #277, opened 2018-06-15
[DR-1649] Share environment configuration between webapp and mseditor

cristiangiagante:
## Background
  
We need to share the environment configuration between the webapp and relay mseditor to call the web api.
  
## Changes done

* Add module dependency to mseditor in order to share environment configs

* Add environment to mseditor bundle script in gulpfile

## Pending to be done
  
Rebase on latest mseditor build version from cdn
  
## Notes
  
We need to define an empty module with the same name than webapp module due to mseditor is a separate angular app, then, we inject environment configuration in this module following the same behavior that webapp originally does. This allow us to share this configuration to the mseditor app.
 
## Demo
 
image


src/wwwroot/template-editor/adapterService.js
```diff
@@ -1,6 +1,9 @@
 (function () {
 
   'use strict';
+  // Add dopplerRelay module to take environment configuration constant
+  angular
+    .module("mseditor").requires.push('dopplerRelay');
```
andresmoschini: How does it work?

src/wwwroot/template-editor/adapterService.js
```diff
@@ -9,13 +12,15 @@
   adapterService.$inject = [
     '$http',
     '$q',
-    '$window'
+    '$window',
+    'RELAY_CONFIG'
```
andresmoschini: is it not necessary to indicate a kind of namespace? or something like that?

src/wwwroot/template-editor/index.html
```diff
@@ -24,10 +24,16 @@
...
       </div>
     </mseditor-loader>
+    
   <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build349/scripts/lib.min.js"></script>
-  <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build349/scripts/app.min.js"></script>
-  <!-- Inject the environment module dependency in build process for template editor-->
+  <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build349/app.min.js"></script>
+  <script>
+    // Create dopplerRelay module out of the scope of webapp to get environment constants
+    angular.module('dopplerRelay', []);
```
andresmoschini: Is it not possible to include it in the bundle?

src/gulpfile.js
```diff
@@ -271,6 +271,7 @@ gulp.task('build-mseditor', ['build-scripts-template-editor', 'add-revision-numb
 
 gulp.task('build-scripts-template-editor', function () {
   return gulp.src([
+    paths.app + '/env/' + process.env.NODE_ENV + '.js',
```
andresmoschini: I understand that you were having issues with scripts sorting inside the bundle, what happens if you add env entry after *.js entry?

src/wwwroot/template-editor/adapterService.js
cristiangiagante: This adds a new module to the current scope of another one in runtime. We use this to prevent override the initial bootstrapped modules from mseditor.

We can't use the next to define it due to this override all the previous bootstrapped modules in mseditor application:
```js
angular.module("mseditor", [dopplerRelay])
```

andresmoschini: Thanks for the clarification!

src/wwwroot/template-editor/adapterService.js
cristiangiagante: Not, due to constants providers works as the same as a service, it is enabled to the entire app once the module that define it is included in the angular app scope.

src/gulpfile.js
cristiangiagante: Oh that was a mistake, the order is the same as the declared order.

src/wwwroot/template-editor/adapterService.js
andresmoschini: In that case, there may not be any advantage in having _dopplerRelay_ module and this could be a better solution: https://github.com/MakingSense/MSEditor/blob/6a2a90d6f442184d93fabdb539540819c82a303f/refactor/gulpfile.js#L481-L488

src/gulpfile.js
andresmoschini: In that case, you could include environment at the end, and you could avoid inline script in the HTML file.

cristiangiagante: Hi andres, you mean to move:
```js
<script>
  // Create dopplerRelay module out of the scope of webapp to get environment constants
  angular.module('dopplerRelay', []);
</script>
```
inside of serviceAdapter at the top and move the js file with the constant definition to the bottom?

andresmoschini: Righ, inside the serviceAdapter or another .js file

src/wwwroot/template-editor/adapterService.js
andresmoschini: I do not understand, I will call you.

cristiangiagante: I will add this approach like a TODO task

cristiangiagante:
Please check my latest changes @andresmoschini I move all environment logic to envConfiguration.js, now we don't use inline scripts in index.html and the env configs are separated from adapterService

src/gulpfile.js
```diff
@@ -271,12 +271,14 @@ gulp.task('build-mseditor', ['build-scripts-template-editor', 'add-revision-numb
...
+    paths.app + '/template-editor/*.js',
+    // TODO: Refactor this config files to json and inject in dopplerRelay and mseditor modules a
+    // configuration module with ng-constant
+    // Reference: https://github.com/MakingSense/MSEditor/blob/6a2a90d6f442184d93fabdb539540819c82a303f/refactor/gulpfile.js#L481-L488
+    paths.app + '/env/' + process.env.NODE_ENV + '.js'
   ])
   .pipe(concat('relay-editor.js'))
-  .pipe(uglify({
-    mangle: false
-  }))
```
andresmoschini: Why did you remove the uglifying?

src/wwwroot/template-editor/index.html
```diff
@@ -26,9 +26,11 @@
     </mseditor-loader>
   <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build515/scripts/lib.min.js"></script>
   <script src="//cdn.fromdoppler.com/mseditor/v1.0.0-build515/scripts/app.min.js"></script>
-  <!-- Inject the environment module dependency in build process for template editor-->
+  <!-- Inject the environment constant RELAY_CONFIG definition dependency in the build process -->
```
andresmoschini: Do you think that it is possible remove this comment?

cristiangiagante: Yea!

src/gulpfile.js
cristiangiagante: 👎 Mistake

cristiangiagante:
Unneccesary comments removed, uglify was added again @andresmoschini

andresmoschini:
:shipit:

PR #285, opened 2018-06-29
Update es-translation.js

andresmoschini:
Hi @mvillaseco, in the end, it was a real bug. It should be fixed now.


andresmoschini:
I will merge it anyway because I have already applied the change in prod, the error seems to be related to bower

PR #292, opened 2018-07-20
[DR-1675] Change get campaign response in adapter

mvillaseco:
# Overview
 Fix issue with return response from adapter

### DIFF
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -396,8 +396,9 @@
           id: id,
           type: 'campaign',
           name: response.data.name,
+          html: response.data.html,
           attributes: response.data.mseditor ? response.data.mseditor.attributes : null,
-          innerHTML: response.data.mseditor ? response.data.mseditor.innerHTML : null,
+          settings: response.data.mseditor ? response.data.mseditor.settings : null,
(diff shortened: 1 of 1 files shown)


src/wwwroot/template-editor/adapterService.js
```diff
@@ -396,9 +396,10 @@
           id: id,
           type: 'campaign',
           name: response.data.name,
+          html: response.data.html,
           attributes: response.data.mseditor ? response.data.mseditor.attributes : null,
-          innerHTML: response.data.mseditor ? response.data.mseditor.innerHTML : null,
-          children: response.data.mseditor ? response.data.mseditor.children : null
+          settings: response.data.mseditor ? response.data.mseditor.settings : null,
+          children: response.data.mseditor ? JSON.parse(response.data.mseditor.children) : null
```
andresmoschini: It should not be necessary.

cristiangiagante: I guess it depends how to children is returned from backend.

andresmoschini: Right, if it is necessary to parse a JSON, there is something wrong.

andresmoschini:
Nice! this PR is really small now.

PR #296, opened 2018-08-08
Dr-1678 allow access to mseditor

mvillaseco:
# Overview
 Add UI access to MSEditor in certain users.

# Changes:
 - Modify template save actions.
 - Add tests
 - Prepare markup to the future html raw option using grid.

# In Progress
 Still missing the new token after refresh


src/spec/template_spec.js
```diff
@@ -335,4 +335,35 @@ describe('Template Page', () => {
...
+
+    // Act
+    browser.get('/#/templates/new');
+
+    // Assert
+    expect(templatePage.isTemplateHtmlRawInputDisplayed()).toBe(true);
+    expect(templatePage.isTemplateMsEditorInputDisplayed()).toBe(false);
+  });
+
+  it('should show the mseditor button when the token forceMseditor is not null', () => {
```
andresmoschini: I think that `forceMseditor is true` or `forceMseditor is truly` is better than `forceMseditor is not null`

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -66,7 +69,11 @@
         id: $scope.template.id
       }).then(function (result) {
         $scope.template.id = result;
-        $location.path('/templates');
+        if(url == "mseditor") {
+          location.href = location.origin + '/template-editor/?idCampaign=' + result;
+        } else {
+          $location.path('/templates');
+        }
```
andresmoschini: I do not like these conditions, `url` name is not clear. What do you think about keeping `save` method private and add another two methods for the different behaviors? call me if you need more details.

src/wwwroot/partials/templates/template.html
```diff
@@ -54,19 +54,45 @@ <h4 ng-show="template.id">{{'templates_information_subtitle' | translate}}{{temp
...
+                  name="uContent"
+                  validation-errors
+                  ng-required="{{!forceMsEditor}}"
+                  ng-model="template.content"
+                  class="input width--full"></textarea>
+      </div>
+      <div class="grid templates-selector--container" ng-show="{{forceMsEditor}}">
+        <div class="templates-mseditor--container">
+          <label>{{ 'templates_edition_mseditor_label' | translate }}</label> 
+          <a class="button-mseditor--selector" ng-click="save('mseditor');submitted = true">
```
andresmoschini: What is `submitted = true`?  please consider a new method in place of using _save_

andresmoschini:
Nice!

PR #297, opened 2018-08-09
[DR-1704] Add cursor pointer in new mseditor button template option

mvillaseco:
# Overview

Add cursor pointer in mseditor option for templates.


andresmoschini:
👍

PR #298, opened 2018-08-10
[DR-1703] Use classic template editor for legacy templates in featured MSEditor accounts

cristiangiagante:
## Background
  
We need to use classic template editor in legacy templates for accounts with MSEditor feature enabled.
We need to be able to save MSEditor template content
  
## Changes done
  
Update Template controller and related partial view to handle Legacy templates

Update Template service to save MSEditor templates and handle legacy templates

Tests added
  
## Pending to be done
  
N/A
  
## Notes
  
Now, when a MSEditor template type is open to edit is not being called anymore the GET /.../templates/{id}/body WEB API call.

### DIFF
--- a/src/wwwroot/controllers/TemplateCtrl.js
+++ b/src/wwwroot/controllers/TemplateCtrl.js
@@ -9,10 +9,11 @@
     '$scope',
     'templates',
     '$routeParams',
-    '$location'
+    '$location',
+    '$rootScope'
   ];
 
-  function TemplateCtrl($scope, templates, $routeParams, $location) {
+  function TemplateCtrl($scope, templates, $routeParams, $location, $rootScope) {
 
     $scope.template = {
       id: "",
@@ -25,12 +26,15 @@
 
     $scope.loadInProgress = false;
     $scope.saveInProgress = false;
+    $scope.isHtmlTemplate = true;
 
-    $scope.saveHtmlRaw = saveHtmlRaw;
+    $scope.saveTemplate = saveTemplate;
     $scope.saveAndEditWithMseditor = saveAndEditWithMseditor;
 
     if ($routeParams["templateId"]) {
       load($routeParams["templateId"]);
+    } else {
+      $scope.isHtmlTemplate = !$rootScope.forceMsEditor;
     }
 
     /*************************/
@@ -39,14 +43,20 @@
 
     function load(templateId) {
       $scope.loadInProgress = true;
-
-      return templates.getTemplateWithBody(templateId).then(function (result) {
-        $scope.template.fromName = result.from_name;
-        $scope.template.fromEmail = result.from_email;
-        $scope.template.subject = result.subject;
-        $scope.template.name = result.name;
-        $scope.template.id = result.id;
-        $scope.template.content = result.body;
+      return templates.getTemplate(templateId).then(function (template) {
+        $scope.template.fromName = template.from_name;
+        $scope.template.fromEmail = template.from_email;
+        $scope.template.subject = template.subject;
+        $scope.template.name = template.name;
+        $scope.template.id = template.id;
+        $scope.isHtmlTemplate = template.bodyType == "rawHtml" || (template.bodyType == "empty" && !$rootScope.forceMsEditor);
+        if (!$scope.isHtmlTemplate) {
+          $scope.template.content = null;
+        } else {
+          return templates.getTemplateBody(templateId).then(function(templateBody) {
+            $scope.template.content = templateBody.html;
+          });
+        }
       }).finally(function () {
         $scope.loadInProgress = false;
       });
@@ -54,23 +64,38 @@
 
     function save() {
       $scope.saveInProgress = true;
-      return templates.save({
+      var templateModel = {
         from_name: $scope.template.fromName,
         from_email: $scope.template.fromEmail,
         subject: $scope.template.subject,
         body: $scope.template.content,
         name: $scope.template.name,
         id: $scope.template.id
+      }
+      var isCreating = !$scope.template.id;
+      var saveTemplatePromise = isCreating ? templates.createTemplate(templateModel) : templates.editTemplate(templateModel);
+      return saveTemplatePromise.then(function(templateId) {
+        if ($scope.isHtmlTemplate) {
+          return templates.editTemplateBody(templateId, templateModel.name, templateModel.body).then(function() {
+            return templateId;
+          })
+          .catch(function (reason) {
+            // TODO: Show a special message error when template was update but body fails
+            return $q.reject(reason);
+          });
+        } else {
+          return templateId;
+        }
       }).finally(function () {
         $scope.saveInProgress = false;
       });
     }
 
-    function saveHtmlRaw() {
+    function saveTemplate() {
       if ($scope.form.$invalid) {
         return;
       }
-      return save().then(function (result) {
+      save().then(function () {
         $location.path('/templates');
       });
     }
@@ -83,8 +108,8 @@
       if ($scope.form.$invalid) {
         return;
       }
-      return save().then(function (result) {
-          location.href = location.origin + '/template-editor/?idCampaign=' + result;
+      save().then(function(templateId) {
+        location.href = location.origin + '/template-editor/?idCampaign=' + templateId;
       });
     }
   }
--- a/src/wwwroot/services/templates.js
+++ b/src/wwwroot/services/templates.js
@@ -17,8 +17,11 @@
 
     var templatesService = {
       getAllData: getAllData,
-      getTemplateWithBody: getTemplateWithBody,
-      save: save,
+      getTemplate: getTemplate,
+      getTemplateBody: getTemplateBody,
+      createTemplate: createTemplate,
+      editTemplate: editTemplate,
+      editTemplateBody: editTemplateBody,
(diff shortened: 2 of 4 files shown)


src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -25,22 +26,60 @@
...
 
-    $scope.saveHtmlRaw = saveHtmlRaw;
+    $scope.save = save;
     $scope.saveAndEditWithMseditor = saveAndEditWithMseditor;
 
     if ($routeParams["templateId"]) {
       load($routeParams["templateId"]);
     }
+    else {
+      $scope.isLegacyTemplate = !$rootScope.forceMsEditor;
```
andresmoschini: I do not like the name, what do you think about `$scope.isHtmlTemplate`? (because it is not legacy, is another option)

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -25,22 +26,60 @@
...
+    $scope.isLegacyTemplate = true;
 
-    $scope.saveHtmlRaw = saveHtmlRaw;
+    $scope.save = save;
     $scope.saveAndEditWithMseditor = saveAndEditWithMseditor;
 
     if ($routeParams["templateId"]) {
       load($routeParams["templateId"]);
     }
+    else {
```
andresmoschini: (very minor) What do you think about moving the _else_ to the same line than the previous curly brace?

```
} else {
```

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -25,22 +26,60 @@
...
+            $scope.isLegacyTemplate = false;
+            break;
+          case "mseditor":
+            $scope.isLegacyTemplate = false;
+            break;
+          case "rawHtml":
+            $scope.isLegacyTemplate = true;
+            break;
+        }
+        if($rootScope.forceMsEditor && !$scope.isLegacyTemplate)
```
andresmoschini: Why do you need to test _forceMsEditor_ value here?

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -25,22 +26,60 @@
...
     /*************************/
 
     function load(templateId) {
       $scope.loadInProgress = true;
-
-      return templates.getTemplateWithBody(templateId).then(function (result) {
+      return templates.getTemplate(templateId).then(function (template) {
+        switch (template.bodyType) {
+          case "empty":
+            $scope.isLegacyTemplate = false;
```
andresmoschini: When _bodyType_ is empty, the behavior should depends on _$rootScope.forceMsEditor_ value

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -25,22 +26,60 @@
...
+          };
+        }
+        else {
+        return templates.getTemplateBody(templateId).then(function (body) {
+          return {
+            from_name: template.from_name,
+            from_email: template.from_email,
+            subject: template.subject,
+            id: template.id,
+            name: body.name,
```
andresmoschini: I think that it is better using _template.name_ in place of _body.name_

Also, there should be a way to reuse the code. Maybe using a kind of _noop_ promise? Let me know if you need help on this.

src/wwwroot/services/templates.js
```diff
@@ -80,17 +81,17 @@
       });
     }
 
-    function save(templateModel) {
+    function save(templateModel, isLegacyTemplate) {
```
andresmoschini: In general boolean parameters are not a good idea, what do you think about splitting _save_ in the same way that you did with _getTemplateWithBody_?

andresmoschini:
Great work! I am proposing some refactoring, let's talk on Monday if you need

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: 👍

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: Yes of course, thx

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: You are rigth, I forgot that, thanks andres.

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: 👍 Ok, let me check it, thx!

src/wwwroot/services/templates.js
cristiangiagante: Ok I will check this method to split it 👍

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: Sorry this is not neccessary, I will remove it.

cristiangiagante:
Thanks @andresmoschini. Could you please review  my lastest changes?

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -39,22 +43,46 @@
...
+        }
+        return {
+          from_name: template.from_name,
+          from_email: template.from_email,
+          subject: template.subject,
+          id: template.id,
+          name: template.name,
+          body: $scope.isHtmlTemplate? templates.getTemplateBody(templateId) : null
+        };
+      }).then(function (result) {
```
andresmoschini: Why did you choose using two _then_?

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -39,22 +43,46 @@
...
         $scope.template.fromName = result.from_name;
         $scope.template.fromEmail = result.from_email;
         $scope.template.subject = result.subject;
         $scope.template.name = result.name;
         $scope.template.id = result.id;
-        $scope.template.content = result.body;
+        return result.body? result.body.then(function(body) {
+          return body.html;
+        }) : null;
+      }).then(function(html) {
```
andresmoschini: I do not understand this code.

```js
        return result.body? result.body.then(function(body) {
          return body.html;
        }) : null;
```

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -39,22 +43,46 @@
...
       }).finally(function () {
         $scope.loadInProgress = false;
       });
     }
 
-    function save() {
+    function saveTemplate() {
       $scope.saveInProgress = true;
-      return templates.save({
+      var saveTreatment = $scope.isHtmlTemplate? templates.saveHtml : templates.saveMSEditor;
```
andresmoschini: I would not differentiate between HTML and MSEditor at template service level, in my opinion, it is better to split saving Template and Template content. If your opinion is different we can discuss it.

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: To return only the value from getTemplateBody's promise and assign it in $scope.template.content

andresmoschini: ok, but why did you split it?

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: result.body is null in the case where the template type is MSEditor, then if result.body is truly I will get a promise to get template body for html templates.

src/wwwroot/controllers/TemplateCtrl.js
cristiangiagante: Please let me know when we can talk about it. Thanks @andresmoschini

cristiangiagante:
Please @mvillaseco, @andresmoschini take a look to my lastest changes, thanks

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -39,39 +43,56 @@
...
+          .catch(function (reason) {
+            // TODO: Show a special message error when template was update but body fails
+            return $q.reject(reason);
+          });
+        } else {
+          return templateId;
+        }
+        
+      }).then(function (result) {
         $location.path('/templates');
```
andresmoschini: It is being executed even when we save to continue editing with MSEditor, please apply a little refactoring here to fix that.

cristiangiagante: 👍

src/wwwroot/controllers/TemplateCtrl.js
```diff
@@ -39,38 +43,57 @@
...
-      return templates.save({
+      var templateModel = {
         from_name: $scope.template.fromName,
         from_email: $scope.template.fromEmail,
         subject: $scope.template.subject,
         body: $scope.template.content,
         name: $scope.template.name,
         id: $scope.template.id
-      }).finally(function () {
-        $scope.saveInProgress = false;
```
andresmoschini: It seems that you have lost this _finally_ block.

cristiangiagante: Yes, sorry thx for the catch

andresmoschini:
Good job!

PR #302, opened 2018-08-27
[DR-1686] Get image gallery mocked.

mvillaseco:
# Overview
In progress advance for the adapter, still need some tests.

### DIFF
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -153,6 +153,7 @@
     function getImagesForCampaign(start, howMany, query, sortingCriteria) {
       traceAdapterCall(getImagesForCampaign, arguments);
       var page = Math.trunc(start / howMany) + 1;
+      var order = sortingCriteria ? (sortingCriteria.isAscending ? 'asc' : 'desc') : null;
       return $http({
         actionDescription: 'Getting image Gallery',
         method: 'GET',
@@ -160,9 +161,9 @@
           query : query || '',
           per_page : howMany,
           page : page,
-          sortingCriteria : sortingCriteria ? sortingCriteria.value : 'DATE',
-          isAscending : sortingCriteria ? sortingCriteria.isAscending : false,
-          onlyImage : true
+          sortBy : sortingCriteria ? sortingCriteria.value : 'date',
+          order : order,
+          onlyImages : true
         },
         url: RELAY_CONFIG.baseUrl + '/accounts/' + loginSession.accountId + '/files',
         headers: {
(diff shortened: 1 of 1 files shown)


src/wwwroot/template-editor/adapterService.js
```diff
@@ -29,8 +29,7 @@
       getPreviewImageWithPlay: getPreviewImageWithPlay,
       getSettings: getSettings,
       refreshMercadoShopsProductServiceToken: refreshMercadoShopsProductServiceToken,
-      getTiendaNubeProducts: getTiendaNubeProducts,
-      overrideUploadFiles: overrideUploadFiles
```
andresmoschini: Is removing this necessary? I mean, _overrideUploadFiles_ seems to be related to uploadings and your changes seems to be related to getting the gallery.

So, we do not know yet if we will need to override some behavior in file uploading.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -71,6 +70,56 @@
...
+     *   fileName: 'testFile.jpg',
+     *   providerName: 'Youtube',
+     *   idCampaign: 1,
+     *   idTemplate: 2
+     * };
+     * 
+     * adapterService.getImagesForCampaign(params);
+     * ```
+     */
+    function getImagesForCampaign(start, howMany, query, sortingCriteria) {
```
andresmoschini: Could you move this block of code to the previous position to see the diff more clearly?

imagen

src/wwwroot/template-editor/adapterService.js
```diff
@@ -71,6 +70,56 @@
...
+      traceAdapterCall(getImagesForCampaign, arguments);
+      var per_page = howMany - start;
+      return $http({
+        actionDescription: 'Getting image Gallery',
+        method: 'GET',
+        params: {
+          query : query || '',
+          per_page : per_page,
+          page : 1,
```
andresmoschini: It is not enough, because the pagination will not work.

src/wwwroot/template-editor/adapterService.js
```diff
@@ -71,6 +70,56 @@
...
+          sortingCriteria : sortingCriteria ? sortingCriteria.value : 'DATE',
+          isAscending : sortingCriteria ? sortingCriteria.isAscending : false,
+          onlyImage : false
```
andresmoschini: I think that onlyImages should be true

src/wwwroot/template-editor/adapterService.js
```diff
@@ -71,6 +70,56 @@
...
+        },
+        url: RELAY_CONFIG.baseUrl + '/accounts/' + loginSession.accountId + '/files',
+        headers: {
+          'Authorization': 'Bearer '+ apiToken
+        }
+      })
+      .then(function(response){
+        var arrayItems = response.data.items.slice();
+        return { data : { 'images' : arrayItems }};
```
andresmoschini: Why did you use slice in place of simply use the previous array?

```
return { data : { 'images' : response.data.items }};
```

PR #312, opened 2018-10-11
[DR-1750] Use links for the thumbnails

mvillaseco:
# Overview
 Use href from the links instead of the obsolete thumbnail property.

### DIFF
--- a/src/wwwroot/template-editor/adapterService.js
+++ b/src/wwwroot/template-editor/adapterService.js
@@ -31,7 +31,7 @@
       refreshMercadoShopsProductServiceToken: refreshMercadoShopsProductServiceToken,
       getTiendaNubeProducts: getTiendaNubeProducts
     };
-    
+
     var langKey = null;
     var loginSession = null;
     var apiToken = null;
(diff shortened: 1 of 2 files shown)


src/wwwroot/template-editor/adapterService.js
```diff
@@ -183,8 +183,20 @@
...
+      .then(function(response) {
+        var items = response.data.items.map(function(item) {
+          return {
+            id : item.id,
+            image : item.image,
+            last_upload : item.last_upload,
+            name : item.name,
+            size : item.size,
+            thumbnailUrl : getHrefFromRel(item, "small"),
+            thumbnailUrl150 : getHrefFromRel(item, "large"),
```
andresmoschini: What do you think about using the full relation name? (in that way it will be reusable for any rel value)

src/wwwroot/template-editor/adapterService.js
```diff
@@ -183,8 +183,20 @@
...
+            url : item.url
```
andresmoschini: Could you also get this URL from a link?

PR #314, opened 2018-10-26
[DR-1766] adaptserService accept array of objects

VicThorX:
Hi !

With help of @andresmoschini I made changes on adapterService to accept array of objects instead array of strings, and send Delete request with the file id.

Any can could review? and help me if I wrong.

PR #315, opened 2018-10-26
[DR-1766] Allow to delete files

VicThorX:
Hi !

With help of @andresmoschini I made changes on adapterService to accept array of objects instead array of strings, and send Delete request with the file id to API.

Could anybody review it? and help me if I wrong.

It replaces #314


andresmoschini:
👍

PR #330, opened 2020-07-08
[DM-464] Add more Billing Information for Doppler SAP

VicThorX:
Hi team!

In this PR I add more billing information when any customer wants to update their plan.
For this, I move some fields to filter some information when is needed, for example, if the country selected is Argentina, we need to show information like "customer type", for more information about these changes read this: [DM-464](https://makingsense.atlassian.net/browse/DM-464?atlOrigin=eyJpIjoiMTU2YjdmNzIxZDgyNGJmNThmMTc2OTU1NzQ2NzFkMzEiLCJwIjoiaiJ9)

Relay WebApp Billing

I still need to do:
- [x] Fix and add tests.
- [x] Finish the `cuit-service` to get information and complete some fields


src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -255,6 +299,28 @@
       }
       redirectToPlanSelection();
     }
+
+    function getCustomerTypes(countryCode) {
```
nreal, line 303: I think that getCustomerTypes should be getConsumerTypes

src/wwwroot/services/settings.js
```diff
@@ -222,5 +223,14 @@
         url: url
       });
     }
+
+    function getCustomerDataByCuit(cuit) {
+      return $http({
+        actionDescription: 'action_getting_customer_data',
+        method: 'GET',
+        url: RELAY_CONFIG.baseUrl + '/customerTax',
```
nreal, line 231: I'm not sure if the URL should be /consumerTax

VicThorX: This will be changed, depend on the [CuitService](https://github.com/FromDoppler/cuit-service)

src/wwwroot/controllers/BillingCtrl.js
VicThorX: Thats true! I don't understand how I got dizzy with it hahaha

src/wwwroot/services/settings.js
andresmoschini: I think that it is enough defined to apply the right URL, see https://github.com/FromDoppler/cuit-service/blob/master/CuitService/Controllers/TaxInfoController.cs#L18

What do you think about 
```suggestion
        url: `https://cuitservice.fromdoppler.com/taxinfo/by-cuit/${cuit}`
```

We can still define the domain, what do you think?

andresmoschini: And, for the fake service, I purpose `https://fakecuitservice.fromdoppler.net/taxinfo/by-cuit/${cuit}`

PR #334, opened 2020-08-06
fix: handle province.code when is undefined

VicThorX:
This error occurs when a client from a country other than 'Argentina' does not select a Province, the province code I am trying to send is not defined.

I don't put a trinary operator (`?` `:`), because the compiler cry always when I use something like that.

There is a better way to handle undefined?


src/wwwroot/controllers/BillingCtrl.js
```diff
@@ -216,6 +217,12 @@
         fiscalId = vm.dni;
       }
 
+      if (vm.province) {
+        provinceCode = vm.province.code;
+      } else {
+        provinceCode = "99";
+      }
```
andresmoschini, line 224:
```suggestion
      provinceCode = vm.province && vm.province.code || "99";
```

Or

```suggestion
      provinceCode = vm.province ? vm.province.code : "99";
```

PR #346, opened 2021-02-03
[DE-30] Html broken template editor. Added  file server setting

CarlosPintos:
The migration to version v1.0.1-build169 needs a file server setting in order to keep working the social follow component


juanasas (approved):
🚀

PR #352, opened 2021-04-28
[DD-119] Fix Resubscribe UI Texts

VicThorX:
Hi team!

Here I modify all text of resubscribe UI, I did not realize that we have a text section!


src/wwwroot/locales/en-translation.js
```diff
@@ -181,8 +181,11 @@
   "resubscribe_label_reason_input": "Reason:",
```
swidzinskiMS, line 181: This is OK en English but in Spanish should be "Motivo" as the confluence documentation says: https://makingsense.atlassian.net/wiki/spaces/DR/pages/775389311/Rehabilitar+emails+bloqueados

PR #356, opened 2021-05-11
feat: Add Billing Information section

jhoffman-ms:
Add the Billing information section to allow to the user changes the credit card information.

**- User with free account:**

image

**- User with bank transfer payment:**

image

**- User with credit card payment:**

image

when the user clicks in the "Cambiar tarjeta" the application shows the following page:

image


andresmoschini:
Thanks @jhoffman-ms!

I think that I will not be able to review this.

Maybe @lucianaDilizio, @ms-darianbenito, @rprietoms, @msimionati or @paangaflo could take a look?

src/wwwroot/controllers/MyBillingInformationCtrl.js
```diff
@@ -0,0 +1,194 @@
...
+
+    function getMaskSecCodByBrand(cardBrand){
+      return secCodeMasksByBrand[cardBrand];
+    }
+
+    function getMaskByBrand(cardBrand){
+      return creditCardMasksByBrand[cardBrand];
+    }
+
+    function getCreditCardBrand(creditCardNumber) {
```
msimionati, line 125: When the billing information is loaded and the user has a credit card, the credit card number is returned obfuscated and the result will be unknown, right?
Now as the api returns the cardBrand, this field should be used to identify the brand, right?

src/wwwroot/locales/en-translation.js
```diff
@@ -511,5 +511,22 @@
...
+  "resubscribe_too_many_request_error": "You have reached your daily reactivation limit.",
+  "submenu_my_billing_information":"Billing Information",
+  "my_billing_information_text": "Billing Information",
+  "my_billing_information_subtitle": "The service will be charged to the credit card listed here. If you wish, you can <b>change the card.</b>",
+  "my_billing_information_change_credit_card": "Change Card",
+  "my_billing_information_back_button": "Back",
+  "my_billing_information_save_button": "Save",
+  "my_billing_information_payment_method_title": "Payment method",
+  "my_billing_information_credit_card_option": "Credit Card",
+  "my_billing_information_card_holder_label": "Cardholder",
```
msimionati, line 523: Is it okay to be like this "Cardholder"?, shouldn't  have space? 
```suggestion
  "my_billing_information_card_holder_label": "Card holder",

File: src/wwwroot/controllers/MyBillingInformationCtrl.js [hunk 3]
```diff
@@ -0,0 +1,193 @@
...
+      vm.cc.parsedCcNumber = utils.replaceAllCharsExceptLast4(vm.cc.number);
+      vm.secCode.ParsedNumber = utils.replaceAllCharsExceptLast4(vm.secCode.number);
+      vm.viewExpDate = form.expDate.$viewValue;
+
+      var paymentMethod = {
+        creditCard: {
+          cardNumber: vm.cc.number,
+          verificationCode: vm.secCode.number,
+          expiryDate: vm.expDate,
+          cardHoldersName: vm.cardHolder
```
darianbenito, line 181: I talked with @jhoffman-ms about including client-side validations using the same server-side validations but It is ok not to include that to be consistent with other Relay webapp sections

msimionati (dismissed):

src/wwwroot/controllers/MyBillingInformationCtrl.js
jhoffman-ms: This method is used only when the user changes the credit card

jhoffman-ms: This method is used only when the user changes the credit card.

src/wwwroot/locales/en-translation.js
jhoffman-ms: yes, is ok: ```"Cardholder"```

nreal:
👍

PR #358, opened 2021-05-18
[DD-483] update QA and INT relay settings

RodrigoPereyraDiaz:
Now all Relay service, APIS, and Workers are in our testing _Swarm_
This PR is for configuring webapp for testing environment to use that instances instead that are on QA windows server

### DIFF
--- a/src/wwwroot/env/INT.js
+++ b/src/wwwroot/env/INT.js
@@ -1,6 +1,6 @@
 ﻿angular.module("dopplerRelay").constant("RELAY_CONFIG", {
-  baseUrl: "https://dopplerrelayint.makingsense.com:4443",
+  baseUrl: "https://int-api-relay.fromdoppler.net",
   cuitServiceBaseUrl: "https://apisint.fromdoppler.net/cuit",
-  hostSmtp: "dopplerrelayint.makingsense.com",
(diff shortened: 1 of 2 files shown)


src/wwwroot/env/INT.js
```diff
@@ -1,6 +1,6 @@
 ﻿angular.module("dopplerRelay").constant("RELAY_CONFIG", {
-  baseUrl: "https://dopplerrelayint.makingsense.com:4443",
+  baseUrl: "https://int.relay.fromdoppler.net",
   cuitServiceBaseUrl: "https://apisint.fromdoppler.net/cuit",
-  hostSmtp: "dopplerrelayint.makingsense.com",
+  hostSmtp: "int.relay.fromdoppler.net",
```
andresmoschini: How does it work? I mean where is defined this name?

andresmoschini: Oh, I understand, it is a new name registered in our DNS.

In general, we are using only one level of subdomain, for example `appint.fromdoppler.net`.

RodrigoPereyraDiaz: Yes es a new name, that is already used in other places from 2 weeks ago to point the new instances running in swarm

PR #359, opened 2021-05-21
Configure Doppler CI

RodrigoPereyraDiaz:
This PR is for enabling Doppler continuous integration.
Also, generate and publish images allowing deploy _Relay Webapp_ as a docker container.

- [x] Fix problem loading chorme-webdriver in CI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,9 +1,15 @@
 pipeline {
     agent any
+    environment {
+        APP_NAME = "relay-webapp"
+        ENVIRONMENT = defineEnvironmentByBranchName()
+    }
     stages {
         stage('Identification') {
             steps {
                 sh 'echo Change request author: ${CHANGE_AUTHOR} '
+                sh 'echo App name: ${APP_NAME} '
+                sh 'echo Environment: ${ENVIRONMENT} '
             }
         }
         stage('Confirmation') {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+
+    return tagMatcher.matches()
+}
+
+// https://stackoverflow.com/questions/56030364/buildingtag-always-returns-false
+// workaround https://issues.jenkins-ci.org/browse/JENKINS-55987
+// TODO: read this value from Jenkins provided metadata
+def String readCurrentTag() {
+    return sh(returnStdout: true, script: "git describe --tags --match v?*.?*.?* --abbrev=0 --exact-match || echo ''").trim()
+}
```
andresmoschini: Now we are using a simpler version that [seems to work better](https://stackoverflow.com/questions/56030364/buildingtag-always-returns-false#answer-65483072):

https://github.com/FromDoppler/hello-microservice/blob/92e58b85074afda427cba2ba7ca5e98a5c3c9740/Jenkinsfile#L102-L107

Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+    post { 
+        cleanup { 
+            cleanWs()
+            dir("${env.WORKSPACE}@tmp") {
+                deleteDir()
+            }
+        }
+    }
```
andresmoschini, line 99: Is it necessary?

andresmoschini:
❤️

Jenkinsfile
RodrigoPereyraDiaz: We are doing the same [post in other projects like doppler-webapp](https://github.com/FromDoppler/doppler-webapp/blob/1acd69f5179ce8edcdac279617dd162801e0b5f8/Jenkinsfile#L104)

andresmoschini: In the web app, it is necessary because we are building the files out of the docker image.

In general, we are trying to avoid doing that.

PR #360, opened 2021-05-21
chore: avoid issues with MINGW and Git Bash in sh scripts

RodrigoPereyraDiaz:


andresmoschini:
👍

src/build-n-publish.sh
```diff
@@ -12,6 +12,12 @@ set -e
 # reference: http://www.ostricher.com/2014/10/the-right-way-to-get-the-directory-of-a-bash-script/
 cd $(dirname $0)
```
andresmoschini, line 13: I know that it is out of the scope of this PR, but maybe this comment could be useful for you.

It could be useful to add ShellCheck CI step (there is also a VS Code extension). You could detect issues like this one (it would fail when the folder name has spaces):

```suggestion
cd "$(dirname "$0")"
```

RodrigoPereyraDiaz: Will be good add _ShellCheck_, a task for another PR :)

PR #368, opened 2022-08-04
Fix continues delivery

RodrigoPereyraDiaz:
We are detected that the environment is wrong configured for INT and should be QA for the default branch to follow the continuous delivery conventions

This will allow use the web app in INT and QA environment

Fix: https://makingsense.atlassian.net/browse/DD-978


ffernandez4 (approved):
🚀

